PR #3, opened 2016-11-15
- improve window detection.

devmondo:
- improve window detection.
- improves element's id content.


devmondo:
- improve window detection.
- improves element's id content.

PR #29, opened 2016-12-22
Stylus plugin

KonstantinKai:


unlight:
So, fusebox is going to be all-in-one tool like webpack? It is not nodish way 👎

Shepless:
@unlight it is like this for now, but in the very near future we will be looking to split plugins out from the main package. FuseBox is still a very young project - much work to do 😄

Issue #31, opened 2016-12-22
Banner Plugin


dogancelik:
Can we add a _BannerPlugin_, I'm writing a Greasemonkey / Tampermonkey script and `// ==UserScript==` has to be at top of the file

https://webpack.github.io/docs/list-of-plugins.html#bannerplugin

![untitled](https://cloud.githubusercontent.com/assets/486818/21444155/6f7831e6-c8b3-11e6-8960-7e63f46dcace.png)

Or is there another way to do this with Fusebox?

Shepless:
@dogancelik @nchanged I have put up a proposal PR of how we can add the functionality in (along with a basic BannerPlugin too). Let me know if the proposal is ok.

nchanged:
Thanks @Shepless it has been [merged](https://github.com/fuse-box/fuse-box#bannerplugin), good job!

PR #32, opened 2016-12-23
Proposal: Pre/PostBundle() Plugin Lifecycle Hooks

Shepless:
Requested from issue #31 

### Changes
- Adds `preBundle()` and `postBundle()` plugin hooks
- Adds very basic `BannerPlugin`

### Notes
I'm not sure if `postBundle()`is very useful. Originally my hope was to maybe use this hook for minification of the entire bundle, but I don't think thats possible. Happy to remove if it adds no value - let me know.


nchanged:
Good stuff!

PR #39, opened 2016-12-24
rename css and html plugin file names

vanthiyathevan:
to match the class name they export. This will make all plugins have same fileName pattern.

export plugins from `plugins/index.ts`


nchanged:
Sorry, it took so long, I was thinking and thinking about it.
Let's keep plugins in index.ts for now. 

We will change it anyway, as soon as we figure out how to deal with plugins.

PR #50, opened 2016-12-29
Verify that PathMaster exists before trying to use it

adrenalin:
Attribute `pm` does not exist always and building crashes because it is missing when chaining several node modules. Here is a patch that passes the tests.


nchanged:
Good job! Thanks

adrenalin:
Off-topic: path resolving has also some similar issues, I will dig into it now. I am trying to import a working React project from Webkit so it gives a nice test ground for a project that is chained heavily. I am absolutely loving the speed of fuse-box!

nchanged:
@adrenalin thanks! That's great that you have some heavy project to test. Fill me in if you get into something.

PR #59, opened 2017-01-02
should fix PostCSS plugin from not resolving the transform/task

matthiasak:
So I was using the PostCSS plugin and found that the CSS plugin was not receiving any contents - the Promise from the `transform()` was not resolving and just timing out. Turns out there was just a small fix: the promise being returned by the PostCSS's `transform()` was never resolving :) fortunately I just had to unwrap the inner call to `postcss.process()` and return that Promise.


matthiasak:
note, here's the API docs for PostCSS's `process()`: http://api.postcss.org/Processor.html#process

PR #66, opened 2017-01-04
Class properties (e.g. stage-2) do not have a callee

adrenalin:
Class properties do not have a callee, but they are handled (apparently) as CallExpression. E.g. this would fail with `stage-2` babel preset

```
class Foo extends React.Component {
  state = {
    foo: 'bar'
  }
}
```

Patched by checking explicitly that callee exists in the preflight check of the node


nchanged:
Good job!

Thanks

PR #69, opened 2017-01-04
Node.js internal packages should be handled

adrenalin:
This is a patch that allows using libraries that use e.g. zlib on server side, but work without on browsers. There is a bunch of node.js internal packages that may depend on the system configuration and those should be taken into account automatically instead of using empty return values of the current `assets/lib/*.js`

This solution is not yet full as those packages should be served on server side normally.

Sorry for the whitespace issues, Atom strips the trailing whitespace automatically.

### DIFF
--- a/src/FuseBox.ts
+++ b/src/FuseBox.ts
@@ -81,6 +81,10 @@ export class FuseBox {
             this.context.ignoreGlobal = opts.ignoreGlobal;
         }
 
+        if (opts.isServer) {
+          this.context.isServer = true;
+        }
+
         if (opts.outFile) {
             this.context.outFile = opts.outFile;
         }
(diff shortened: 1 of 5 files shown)


nchanged:
Thanks for your idea, but I can't merge it. The thing is that it __should__ be bundled. Otherwise, bundle won't work in browser.  Please check this [PR](https://github.com/fuse-box/fuse-box/pull/71)

A simple solution is to add it [here](https://github.com/fuse-box/fuse-box/tree/master/assets/libs) 

JSPM works the same way. __zlib__ was simply added to the list https://github.com/fuse-box/fuse-box/blob/master/assets/libs/zlib/index.js

adrenalin:
This sounds like patching over patching with a lot of redundancy needed to support all the missing packages, but ok, as you wish. E.g. `tty` is currently breaking [debug](https://www.npmjs.com/package/debug) because it "exists" but is broken in the eyes of the package, because it doesn't have any of the required methods after the check for existence has been done.

nchanged:
The issue is a bit different. It bugs up on cache

Cuz this [works](https://github.com/fuse-box/fuse-box/blob/master/src-frontend/FuseBox.ts#L174)

nchanged:
@adrenalin you can try fixing the cache instead. Cuz if a package is not in the asset lib, it will still work.

nchanged:
@adrenalin join gitter, i can explain.

PR #95, opened 2017-01-23
Fix issue not able to use hot reload over ssl

erkkaha:
window.location.origin is used instead of window.location.host. This way protocol is included and matching will be done correctly and hot reload will also work over ssl connections.


nchanged:
Good one!

PR #131, opened 2017-02-11
:memo: improve bundling docs

basarat:
- formatting cleanups 
- Mention that bundle inputs are relative to `homeDir` 
- mention that outFiles if used is relative to `process.cwd()`

Thanks! :rose:


nchanged:
Thank you! Docs are getting polished :)

PR #132, opened 2017-02-11
Cleanup developer workflow

basarat:


src/FuseBox.ts
```diff
@@ -81,7 +100,7 @@ export class FuseBox {
             this.context.shim = opts.shim;
         }
 
-        if (opts.standalone !== undefined) {
+        if (opts.standaloneBundle !== undefined) {
```
basarat: probably fixed a :bug: here. `standaloneBundle` isn't documented http://fuse-box.org/ though so its okay :rose:

src/WorkflowContext.ts
```diff
@@ -11,234 +11,80 @@ import { EventEmitter } from "events";
...
     public tsMode = false;
 
-    public loadedTsConfig: any;
-    /**
-     * 
-     * 
-     * @type {string[]}
-     * @memberOf WorkFlowContext
-     */
-    public globals: any;
```
basarat: the `@type` was inconsistent with docs. Adding a type annotation for fusebox options caught it though :rose:

src/FuseBox.ts
nchanged: It will be implemented soon. I removed it earlier for some reason ...

PR #142, opened 2017-02-12
more documentation

basarat:
:rose:


src/plugins/HotReloadPlugin.ts
```diff
@@ -12,7 +12,7 @@ import { Plugin } from "../WorkflowContext";
 export class HotReloadPluginClass implements Plugin {
     public dependencies = ["fusebox-hot-reload"];
     public port: any = "";
-    constructor(opts: any = {}) {
+    constructor(opts: { port?: number | string } = {}) {
```
nchanged: that's pure magic. I love typescript, will use it from now on.
Thanks!

PR #147, opened 2017-02-12
Extended tty definition to support for debug.js

tomitrescak:
debug.js is a package used by many libraries out there and it depends on tty library. Following is a stub implementation of the only tty call needed by debug.js.


nchanged:
Thank you very much!

PR #151, opened 2017-02-13
docs: Correct Home Directory being relatieve to AppRoot

DoumanAsh:
subj.

I also added explicit mention of fuse-box creating directory structure for `Out File`

**UPD:** Sorry, my editor takes care of trailing whitespaces. Let me know if you mind it

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,7 +13,7 @@ FuseBox.init({ /* you config is here */ })
 
 ## Home directory
 
-That's your source folder. It can be an absolute path, Or relative to `process.cwd()`.
+That's your source folder. It can be an absolute path, Or relative to `appRootPath`.
 
 ```js
 FuseBox.init({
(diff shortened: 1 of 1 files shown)


nchanged:
Thanks!

basarat:
> docs: Correct Home Directory being reltive to AppRoot

Don't think it is right. Its never passed through to `ensureUserPath` and stays relative before e.g. being passed to watchTree 

image

And its app root path stuff isn't used in `Arithmetic.ts` either

image

Without being passed to ensureUserPath it stays relative to `process.cwd` :rose:

DoumanAsh:
Now that's strange, we're talking about homeDir, right?
https://github.com/fuse-box/fuse-box/blob/master/src/FuseBox.ts#L67-L69

basarat:
@DoumanAsh I missed that line. You are right! it is relative to `appRootPath` :rose: 

I'm actually glad cause this was the only thing that I thought wasn't `appRootPath` and it stood out in my brain as a sore inconsistency. Its all consistent :heart: :)

PR #154, opened 2017-02-13
Check if imported package has a browser-field and use it

gerhardsletten:
Npm packages [can have a browser-field](https://github.com/defunctzombie/package-browser-field-spec) if it supports both node and browser, Fuse-box should use the this fields.

### DIFF
--- a/src/PathMaster.ts
+++ b/src/PathMaster.ts
@@ -278,19 +278,15 @@ export class PathMaster {
                 // Getting an entry point
                 let entryFile;
                 let entryRoot;
-                if (json.main) {
-                    entryFile = path.join(folder, json.main);
-                } else {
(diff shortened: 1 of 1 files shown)


nchanged:
Tests have failed, please run `mocha path_master.js` to fix it.

gerhardsletten:
@nchanged It was not `path_master.js` but `heavy_node_modules.js` that failed. Btw there is several tests that fail when I run `npm test` locally, but apparently not  on Travis. Do the test require some special setup locally or some global modules or binaries?

nchanged:
That's right https://github.com/fuse-box/fuse-box/blob/master/CONTRIBUTING.md

nchanged:
I will merge you a bit later! Thanks for the PR!

Issue #148, opened 2017-02-12
Hot-reload with redux-provider


gerhardsletten:
Should this work? I have tried to extend the react-example with a typical setup with redux and react-router, but get and error `<Provider> does not support changing `store` on the fly...` when fuse-box tries to do a hot reload.

It also looks like the setting `cache: false` disables the hot-reload.

skellock:
Redux with hot reloading is working for me.  Mostly.  Something is off, and I can't quite put my finger on it.  

Usually you can get around that `<Provider>` warning by making changes to a component deeper down.  But for some reason it's causing the whole display tree to re-render.

If you want to not reload your store everytime, one option is to stick the instance on `global`,  `window`, or someplace equally as horrifying. 

```js
window.storeAndAlsoImGoingToHell = window.storeAndAlsoImGoingToHell || createStore()

// then later in your file
ReactDOM.render(
    <Provider store={window.storeAndAlsoImGoingToHell}>
        <Hello />
    </Provider>
    ,
    document.querySelector(element)
);
```

Also, can confirm, `cache: false` indeed seems to break hot-reload.

nchanged:
Hey guys!
Hot reload works only with cache on. Indeed, that's by design. At least for now. At the very moment, Hot reload flushes everything and reloads the entry point. Yes, there is a room for improvement. 

Let's discuss how can we do it more efficiently.

gerhardsletten:
@skellock Thanks got it working now. Actually I did not need to make the store global to have in preserved through a hot-reload. 

@nchanged Is the rule that the creation of the Provider, Routes show not be in the entry-file? There are probably more then me that are a little unsure of the success criteria's for hot-reload to work.

My test project can be found here, its still some work in progress.. 

[fuse-box-react-ssr](https://github.com/gerhardsletten/fuse-box-react-ssr)

gerhardsletten:
I have also discover that with EnvPlugin, the global var process is not available after a hot-reload, you need to do a import for it to be:

```js
import process from 'process'
console.log(process.env.someVar)
```

nchanged:
@gerhardsletten thanks! that's a real bug. Could you create a separate issue?

nchanged:
Guys, we are cooking HMR plugin support. You will have a full control on which module to reload.

basarat:
> <Provider> does not support changingstoreon the fly...

[...]

Issue #115, opened 2017-02-10
development server not refreshing properly


lionar:
# Development server refresh

When i run the fusebox development server and change my content the following happens:

- On Firefox the development server refreshes the page once properly, after that it does not refresh anymore. 

- On chrome the development server refreshes once properly, after that it keeps appending my content when i use for example document.write, meaning the old content that was previously written stays on the page in chrome.

i used the following configuration:

```js
const fb = require ( "fuse-box" );
const FuseBox = fb.FuseBox;

const fuse = FuseBox.init ( 
{
	homeDir: "./application",
	outFile: "./build/application.js"
} ); 

fuse.devServer ( "> index.ts", 
{
    port: 4444, 
    httpServer: true
} );
```

nchanged:
Welcome @lionar and thanks for submitting the issue.

FuseBox reloads your entry point, therefore everything is repeated. `document.write` will be executed as many times as it was called.

devmondo:
the `HMR` is more powerful now and gives you ability to customize the behavior per requirement, please check this article for more info 
[Rethinking Hot Module Reloading](https://medium.com/@basarat/rethinking-hot-module-reloading-58ce15b5f496#.uw41smbh4)

Issue #98, opened 2017-01-25
devServer API improvements


nchanged:
HMR requires improvement:

* Request a partial reload
* Request flushing on demand (removing modules from cache) using a pattern

devmondo:
#115

basarat:
> Request flushing on demand (removing modules from cache) using a pattern

Think https://github.com/fuse-box/fuse-box/pull/158 will solve it :rose:

PR #158, opened 2017-02-14
HMR plugins

basarat:
⚠️ Work in progress. Will message once done :rose: 

All HMR issues seem to be about not wanting the cache to clear for certain modules:

* https://github.com/fuse-box/fuse-box/issues/148 (`<Provider> does not support changingstoreon the fly...`)
* https://github.com/fuse-box/fuse-box/issues/115 (user wants to exclude the main module)
* https://github.com/fuse-box/fuse-box/issues/98 (Meta issue but only links to #115 right now)


basarat:
Don't merge. Will talk more tomorrow 🌹

nchanged:
Okay!

leon-andria:
Does HMR should be multiple in an application ?
Or they should be chainable ?
It might dangerous to make it just an array

nchanged:
It might, not merging anything now, but we need to have HMR plugin support.

leon-andria:
Of course. HMR is very important.
I just highlight the fact that speed implementation does match with quality and sustainability.  :-)

basarat:
@nchanged just a quick review (don't merge as I need to test it). Main commit is https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb

The way it works: 

* User passes option `hmrStatefulPlugins` to `devServer` and we just send it to `HotModuleReloadPlugin` : https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb#diff-8c475be8a75f3fc4045c0fc95d5c54dfR61 
*  That in turn just passes it to the front-end `fusebox-hot-reload` : https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb#diff-b7e42c8e1abbd7f575cbbe589fce3215R8 
* which then uses it to register a plugin (with essentially itself) https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb#diff-b7e42c8e1abbd7f575cbbe589fce3215R17 for a different behavior (flushing other files except the stateful ones unless some stateful one has changed in which case all files are flushed). 

My suggestion: 

* I remove `addHMRPlugin` and just move the code inline here : https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb#diff-b7e42c8e1abbd7f575cbbe589fce3215R46
* Or I can still leave the code for `addHMRPlugin` and just not use it. 

As always, thanks! :rose:

nchanged:
That's just terrific! thanks!! Just woke up and what a nice surprise!

nchanged:
Would be cool to have one flush method with options. Maybe regexp... or test filter function?! :)

basarat:
> Would be cool to have one flush method with options. Maybe regexp... or test filter function?! :)

Done! :rose: https://github.com/fuse-box/fuse-box/pull/158/commits/69395fcdb03bce94f5372908652a811a4bfc2a32 

My questions in https://github.com/fuse-box/fuse-box/pull/158#issuecomment-279922333 still stand :heart:

nchanged:
Ok. what about a default `hmr-default-plugin` in [here](https://github.com/fuse-box/fuse-box/tree/master/modules) ? And I think you missed the CSS reload. We need tests for that...

basarat:
> And I think you missed the CSS reload. We need tests for that...

Explicitly didn't cater for it https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb#diff-b7e42c8e1abbd7f575cbbe589fce3215R35 I don't see people wanting to preventing the default behavior ( __fsbx_css https://github.com/fuse-box/fuse-box/pull/158/commits/3e87d418b48e77602183aad744499b75ab5508bb#diff-b7e42c8e1abbd7f575cbbe589fce3215R60  ) for CSS files 

> what about a default hmr-default-plugin in here 

Don't understand. I'm available on gitter for the next 1 hour https://gitter.im/basarat if you want to discuss there :heart:

nchanged:
Wow! Is it ready?)

basarat:
Nope. Working on adding devserver support for loading these 🌹

docs/loader-api.md
```diff
@@ -141,3 +141,60 @@ require("~/lib/utils")
...
+    /** 
+     * If true is returned by the plugin
+     *  it means that module change has been handled
+     *  by plugin and no special work is needed by FuseBox
+     **/
+    hmrUpdate?(evt: SourceChangedEvent): boolean;
+}
+
+/** Where */
+type SourceChangedEvent = {
```
Thanood: You've changed `type` to `interface` in `LoaderAPI.ts` - I'd propose to change this line, too. 😃

basarat: I'm scared of people being confused by which interface they need to implement. I know sounds weird but who knows 🌹

nchanged:
Hi! You should probably sync with master =)

basarat:
Will do :rose:

nchanged:
Interesting.. it's not happening in master...

basarat:
> Interesting.. it's not happening in master...

Suspect me touching the loader API has pushed the size over the hard coded limit `5300` ?

nchanged:
Okay, can you increase it a little? But just enough=)

basarat:
@nchanged Ready to merge / release whenever you want :1st_place_medal:  no hurry :heart:

magic

Tested on the fuse-box-reflux example. Thanks for making such an awesome ecosystem! :rose: :heart:

nchanged:
@basarat :fire: That's just astonishing!

nchanged:
Coming in the next release!

basarat:
Note to users : Docs for this feature are published as well : http://fuse-box.org/#loader-plugins :rose:

tomitrescak:
Thanks for the amazing PR! Looking forward to it like a baby to a new rattle ;)
Will you provide an example of how to include React HMR in an app? I looked at docs but to be hones, not too smart from it :/

basarat:
# Once this is released 

@tomitrescak After this is released you would copy paste this (from the docs http://fuse-box.org/#loader-plugins)

```
const registerStatefulModules = (moduleNames: string[]) => FuseBox.addPlugin({
  hmrUpdate: ({ type, path, content }) => {
    if (type === "js") {
      const isModuleNameInPath = (path) => moduleNames.some(name => path.includes(name));

      /** If a stateful module has changed reload the window */
      if (isModuleNameInPath(path)) {
        window.location.reload();
      }

      /** Otherwise flush the other modules */
      FuseBox.flush(function(fileName) {
        return !isModuleNameInPath(fileName);
      });
      /** Patch the module at give path */
      FuseBox.dynamic(path, content);

      /** Re-import / run the mainFile */
      if (FuseBox.mainFile) {
        FuseBox.import(FuseBox.mainFile)
      }

      /** We don't want the default behavior */
      return true;
    }
  }
});

registerStatefulModules(['TestStore', 'actions/index']);
```
into the bottom of this file: https://github.com/fuse-box/fuse-box-ts-react-reflux-seed/blob/0c659e56eca78f8edf8f673c7b55e81385e94043/src/client/store/TestStore.ts 


# Future

Ideally this code should be written in something that is only loaded by the dev server instead of polluting your application.  I was thinking 

```ts
.devServer('>yourApp.tsx', {serverBundle:'>moreCodeThatCustomizesTheServer'})
```

Haven't finalized that idea yet though :rose:

nchanged:
@basarat can this file be at the projects' root? With test for `process.env.NODE_ENV` that would make sense =)

basarat:
> can this file be at the projects' root?

Not quite as the project root is run again and again by 

```ts
/** Re-import / run the mainFile */
      if (FuseBox.mainFile) {
        FuseBox.import(FuseBox.mainFile)
      }
```

So it would register the plugin again and again. Has to be in some stateful file and since we reload the page on a stateful file change its not a problem. Ideally it would be in a server only bundle making it immune to HMR updates :rose:

nchanged:
@basarat but we could blacklist it :-)
Besides people will use to reload their modules without touching an entry point. We just need to make it simple to use. I gave you write access to this repo https://github.com/fuse-box/fuse-box-hmr-examples could put cook some examples for people? ;-)

tomitrescak:
Examples would be great! I'm still not sure what is expected in registerStatefulModules. Do I need to register every single ui component, or just a router, for example. 

So an example of SPA, with (react/mobx/yester) router and couple routes would be peachy;)

basarat:
@tomitrescak https://github.com/fuse-box/fuse-box-ts-react-reflux-seed has been updated. `improveHMR` 👍 

Will work on a yester example next :rose:

basarat:
> Do I need to register every single ui component, or just a router, for example.

No. Just the files containing `state` and the `router`. Hold tight. Example coming :heart:

tomitrescak:
In the seed example , why did you use 'TestStore' module as path when you used 'actions/index's? It seems that correct path would be 'store/TestStore', what am I missing?

basarat:
@tomitrescak as long as *some* portion of the path matches it is okay. https://github.com/fuse-box/fuse-box-ts-react-reflux-seed/blob/df23675337e37013b6d8bb9803de0423400b029a/src/client/hmr.ts#L5 

I should just make it `store` and `actions` :rose:

PR #171, opened 2017-02-14
Rename PostCSS to PostCSSPlugin

7Pass:
So that it's consistent with other plugins. Docs also use the name PostCSSPlugin.

### DIFF
--- a/src/plugins/PostCSSPlugin.ts
+++ b/src/plugins/PostCSSPlugin.ts
@@ -56,6 +56,6 @@ export class PostCSSPluginClass implements Plugin {
     }
 }
 
-export const PostCSS = (processors?: any, opts?: any) => {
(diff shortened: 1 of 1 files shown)


nchanged:
I am not sure we can merge it. There might be people using it already. We need to move it a major release.

basarat:
Should have broken the build as `index.ts` wasn't changed https://github.com/fuse-box/fuse-box/blob/0c4b1c322ca3aa2ad7019b2e77f813e8f9afb81e/src/index.ts#L8

basarat:
if there will be a major version update should probably rename all the other plugins as well https://github.com/fuse-box/fuse-box/blob/0c4b1c322ca3aa2ad7019b2e77f813e8f9afb81e/src/index.ts#L1-L22 :rose:

nchanged:
Cool! Thanks!
Closing this for now.

PR #190, opened 2017-02-16
change brightBlack to reset get neutral color on dark themes

### LINKED ISSUE #173: Bundle 2nd level color is invisible on solarized dark
Here is a fun one, I thought these lines were just empty until this ran on my build server.

image

The color shows up under highlight, and under other themes (like solarized light). It just happens to be the background color for Solarized Dark.

tyrsius:
To fix #173.

Grey colors do not show up on solarized dark (and a few others). But the neutral color is already grey.

Here is solarized dark before the change (with some highlighing to show the invisible text)
image

Here is solarized dark after the change
image

For extra comparison here is Solarized light before and after
image

image

I tested several other themes just to be sure, but since this uses `reset()` it should be safe everywhere.

### DIFF
--- a/src/Log.ts
+++ b/src/Log.ts
@@ -41,14 +41,14 @@ export class Log {
         let bytes = Buffer.byteLength(contents, "utf8");
         let size = prettysize(bytes);
         this.totalSize += bytes;
-        cursor.brightBlack().write(`└──`)
+        cursor.reset().write(`└──`)
             .green().write(` ${collection.cachedName || collection.name}`)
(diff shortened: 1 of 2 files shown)


tyrsius:
I forgot to mention, I tried the other grey/black/bright colors, but they produced poor looking reports. White looked ok, but it seemed to break the intention of the muted color brightBlack is supposed to produce. All the other colors clashed with the green and yellow in the report.

nchanged:
Good job! Thanks a lot!

Issue #167, opened 2017-02-14
UglifyJS Plugin should report on progress and recalculate the bundle size


nchanged:
subj

#153

tyrsius:
This can be closed now

PR #192, opened 2017-02-16
Report the new size and time taken for UglifyJS

tyrsius:
This is for #167, but it is not ready. I want to start a discussion.

This is what the report looks like currently
image

Obviously the lack of a label on the second report is one issue. I'm a little confused on how @nchanged wants to organize things. Things seem sort of abstracted around the `context` object, but the `Log` modules on `context` seems pretty specific to logging `ModuleCollection`, and their isn't another logging module. 

* Do you want a new logging module to abstract logging?
* Do you want `cursor` use directly in the `UglifyJSPluginClass`
* ... something else?

### DIFF
--- a/src/FuseBox.ts
+++ b/src/FuseBox.ts
@@ -222,7 +222,7 @@ export class FuseBox {
                 }
 
             }).then(result => {
-                self.context.log.end();
+                self.context.log.end('FuseBox Bundle');
                 this.triggerEnd();
                 self.context.source.finalize(bundleData);
                 this.triggerPost();
--- a/src/FuseBox.ts
+++ b/src/FuseBox.ts
@@ -222,7 +222,7 @@ export class FuseBox {
                 }
 
             }).then(result => {
-                self.context.log.end('FuseBox Bundle');
+                self.context.log.end();
                 this.triggerEnd();
                 self.context.source.finalize(bundleData);
                 this.triggerPost();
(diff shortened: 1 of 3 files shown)


nchanged:
Great, let's improve the logging class. It makes sense IMO, 
We need to add warnings maybe?!

tyrsius:
Warnings for what?

nchanged:
In general. Just saying. Should we change the output a bit, when uglified. 
____
Size: 52kb (minified)
Time: 677ms
___
And why you have 2 of them?

tyrsius:
I think if they were both labeled beforehand, like `Fuse bundle` and `UglifyJSPlugin bundle`, the (minified) part would be unnecessary, and make the bundle report summary consistent.

For two of them, to stop the first one from happening you would have to know that that `UglifyJSPlugin` was going to run. Even if you wanted to check for built in plugins that would change the bundle, which I think might be hard, the fuse bundle report is still useful info.

nchanged:
https://github.com/fuse-box/fuse-box/blob/master/src/Log.ts#L78

Shall we expose this one? Or make it customisable?

nchanged:
That means that the Log should have `context`.

tyrsius:
That's what I did here, check out the changed files. For getting the label, if you want to make `Log` the generic module for that, I think it needs a generic `echo` function

tyrsius:
Current revision with label, no hyphens. Ready to merge if you like it

image

nchanged:
Terrific job! Thanks!

PR #195, opened 2017-02-16
Fix files not bundled on windows OS when root starts with lower case  drive letter

### LINKED ISSUE #189: Incorrect behavior (no files bundled) on Windows with lowercase drive letter
Hello,
This project seems very promising, we falled in love in less of one day!

We found an issue on Windows machine, that can be replicated running fuse bundler (even the sample apps) in a prompt where _the drive is loaded with a lowercase letter_.
You should be able to create such prompt using a Windows shortcut to cmd.exe and setting the "Start in" field with an absolute path that starts with a lowercase letter (e.g. 'e:\dev').

Strangely, Windows allows this, and then any reference to absolute path in applications becomes lowercase, making some regex in PathMaster to fail (NODE_MODULE).

The issue is quite nasty because difficult to spot (FuseBox in that case will produce an empty bundle without any error).

Thank you for the support on this project!
 
capture
Luciano

devmondo:
Fix files not bundled on windows OS when root starts with lower case  drive letter closes #189

### DIFF
--- a/src/PathMaster.ts
+++ b/src/PathMaster.ts
@@ -9,7 +9,7 @@ import { Config } from "./Config";
  * If a import url isn't relative
  * and only has ascii + @ in the name it is considered a node module
  */
-const NODE_MODULE = /^([a-z@].*)$/;
+const NODE_MODULE = /^[^:]([a-z@].*)$/;
 
 export interface INodeModuleRequire {
     name: string;


src/PathMaster.ts
```diff
@@ -9,7 +9,7 @@ import { Config } from "./Config";
  * If a import url isn't relative
  * and only has ascii + @ in the name it is considered a node module
  */
-const NODE_MODULE = /^([a-z@].*)$/;
+const NODE_MODULE = /^[^:]([a-z@].*)$/;
```
basarat: Thought this match was only for `"foo"` or `"foo/bar"` or `"@types/foo"` etc. 

Not full file paths e.g. it doesn't support `/` either. I can be wrong :rose:

devmondo: @basarat yeah it is failing on linux, but working on windows and the previous one is doing it visa versa

nchanged: @basarat that's some odd bug in here https://github.com/fuse-box/fuse-box/issues/189
I have no idea yet what's happening there, or even what they mean :(

nchanged: @basarat and you are right, that's what it does. Only matching node module names, but somehow C:\ got in there...

basarat: > Only matching node module names, but somehow C:\ got in there...

Think it should be `/^([a-z@]*)$/`. There is an extra dot in `.` in the current `/^([a-z@].*)$/` that is wrong :rose:

nchanged: I think i got. You can't require absolute paths in fuse.
```
 require(someABSPathOnWindows)
```

Or even in Unix... 
that has to be fixed.

nchanged: Hm.. @devmondo can you try that? @basarat thanks!

basarat: fwiw official package name tester : https://github.com/npm/validate-npm-package-name but that has more validation than what you probably want : https://github.com/npm/validate-npm-package-name/blob/ddf73c79e920b59413485ca2ae50a38552156d62/index.js :rose:

devmondo: @basarat `/^([a-z@]*)$/` blows up fuse hahaha the dot is important especially on windows as @nchanged said above 🌹

nchanged: @basarat that's handy. But would it affect the performance?

nchanged: And I think that even if that is fixed, this file is going to go further, breaking other things ...

basarat: Yeah I realize the dot is also for `/` e.g. `@types/foo` or even `foo/something/inside` :(

nchanged:
Clearly this does not work. Let's work on other solutions

lmartorella:
Hi, perhaps the correct regex should be
/^([a-z@](?!:).*)$/

so: start line, then a lower-case letter or @, then negative lookahead for an immediate following colon char.
Thx! L

PR #200, opened 2017-02-17
add expose possibilities

tyrsius:
Currently it is possible to expose a package to the bundle by name. Unfortunately when bundling CommonJS modules this means a wrapper is required to consume them. To understand why, consider this module

```js
exports.handler = function () { }
```

When required, the pre and post module have a different shape.
```js
var lambda = require('pre-bundled-lambda') // { handler: [Function] }
var lambda = require('bundled-module') // { lambda: { handler: [Function] } }
```

This is easily remedied by allowing the `globals` property of the bundle configuration to specify deeper values. I have two solutions, both of which can live together in harmony.

```js
globals: { 'lambda': '*' },
globals: { 'lambda': { 'handler': 'handler' } }
```

The loader API exposes these with this modification
```js
public static expose(obj: any) {
        for (let key in obj) {
            let data = obj[key];
            let exposed = $import(data.pkg);
            if (data.alias === '*') {
                for (let exportKey in exposed) {
                    if (exposed.hasOwnProperty(exportKey)) {
                        __root__[exportKey] = exposed[exportKey];
                    }
                }
            } else if (typeof data.alias === 'object') {
                for (let exportKey in data.alias) {
                    if (data.alias.hasOwnProperty(exportKey)) {
                        __root__[data.alias[exportKey]] = exposed[exportKey];
                    }
                }
            } else {
                __root__[data.alias] = exposed;
            }
        }
    }
```

For our module both formats will expose the `handler` function, but the second pattern allows a bit more control, while also being a little clearer.

I have tried to maintain the structure of the `globals` object.

```js
globals: { 'packageName': 'nameOfExportInBundle' } // old method, nameOfExportInBundle is value
globals: { 'packageName': { 'nameOfExportInPackage': 'nameOfExportInBundle' } // nameOfExportInBundle still in value
globals: { 'packageName': '*' } // * = all exported properties
```


tyrsius:
I think the tests might be a little too strict. The first failing one is looking for size under `5400`, the size is `5452`.  The rest of the failures are all similar, off by less than 50 bytes

nchanged:
Ok, let's see. The loader API is getting bigger and bigger :) will check it soon

nchanged:
Great! Let's increase it, but just enough. And could you be so kind to document the feature?

tyrsius:
Of course, I can add some tests too (in the morning, I'm about to go to sleep). I just wanted to get your thoughts first. Do you want the wildcard and the object format, or just one of them?

nchanged:
I think we can have both of them? Do you think it makes sense?

tyrsius:
They both work, and I like them. You mentioned wanting to keep the API small though, thought you might not want both.

nchanged:
Yeah, I think we can optimise it, it's better to have features, rather than having 50 bytes less ;-)

tyrsius:
Working on optimizing the byte size of these changes, is there an ES compatability target? Does this bundle need to run in IE8? Can I use `Object.keys` and `Array.prototype.forEach`?

nchanged:
@tyrsius typescript does not convert it right?! It's better to keep it compatible with everything. Without polyfils

nchanged:
Btw, uglifyJS uglifies all the local variables nicely. like $pathJoin to one letter.

tyrsius:
No, typescript does not convert or polyfill those. They both require IE9

tyrsius:
Ok, added tests and documentation and optimized the code in the LoaderAPI. Should be ready to go

nchanged:
What about tests on the "default" package without `package` defined. All works as before?

tyrsius:
I'm not sure what you are referring to, but I only added two tests and modified the maximum byte size on the bundle in bundle tests. If there was a test on that, it should be fine.

nchanged:
Good, thanks! But let's keep in the mind that the size of the API matters. We should shrink it down in the future.

tyrsius:
I was focused on shrinking the code I added, but I will spend some more time looking for existing code to shrink. I didn't actually realize when I started this that the entire LoaderAPI is included in the bundle!

nchanged:
@tyrsius that's the problem =) Great feature btw. Now you understand why I mentioned the size =) Features are nice, but 50 bytes here 50 there and the loader is bloated, which we should keep very succinct.

PR #211, opened 2017-02-17
Reverted wrong fix #197

### LINKED ISSUE #197: PostCSS plugin doesn't do anything with opts?
The PostCSS plugin does not pass opts into postcss.  Just seems like an omission.  The opts are essential for implementing CSS modules in PostCSS.

Thanks!

unlight:


nchanged:
@unlight it does not accept params?

unlight:
Check [JS API](https://github.com/postcss/postcss#js-api)
PostCss plugin usually exposed as a [factory method](https://github.com/fliptheweb/postcss-flow-root#usage)
E.g.
```
processors = [
  require('postcss-flow-root')(options),
  require('postcss-plugin-2')(options2),
]
```
Correct call postcss process:
`postcss(processors)` or ``postcss(...processors)`

Incorrect call:
`postcss(this.processors, this.opts)`
`this.opts` acts as plugin here not as plugin.

Options should be passed to `.process(file.contents)`
I'll add commit... 🐤

Issue #213, opened 2017-02-17
Trying to generate React bindings in non-tsx files?


zeh:
Apologies if this is not the right place for questions, but I've started porting a site build system I have from Webpack to Fuse-box and ran into a snag. I would appreciate any input.

The project in question uses TypeScript 2.2 + React + Redux + PostCSS. This is my current attempt at a build script:

```javascript
const fsbx = require("fuse-box");

fsbx.FuseBox.init({
    homeDir: "src/",
    outFile: "./dist2/app.js",
    sourceMap: {
        bundleReference: "app.js.map",
        outFile: "./dist2/app.js.map",
    },
    plugins: [
        fsbx.TypeScriptHelpers(),
    ]
}).devServer(">[ts/main.tsx]");
```

When I run this, I get this message:

```shell
$ node build
16:30:54 : Launching dev server on port 4444
  → Typescript config:  \tsconfig.json
Acorn error: Unexpected token (13:4)
File: D:\<...>\src\ts\data\stores\reducers\search.ts

11      React.createElement("string", null,
12          "):ISearch => ",
13      ,
14          " ",
15          React.createElement(model_1.ISearch, null,
16              "Object.assign(",
17              ", state, ",
```

So it's apparently trying to interpret that file as a React file, and tripping on the comma at line 13. But it's not what the file is; it's just a reducer to a Redux store. This is its full contents:

```typescript
import { Action, handleActions } from "redux-actions";
import { ISearch } from "./../model";
import {
	CLEAR_SEARCH_TEXT,
	SET_SEARCH_TEXT,
} from "./../actions";

const initialState:ISearch = <ISearch>{
	text: "",
};

export default handleActions<ISearch, any>({
	[SET_SEARCH_TEXT]: (state:ISearch, action:Action<string>):ISearch => {
		return <ISearch>Object.assign({}, state, {
			text: action.payload,
		});
	},
	[CLEAR_SEARCH_TEXT]: (state:ISearch, action:Action<void>):ISearch => {
		return <ISearch>Object.assign({}, state, {
			text: "",
		});
	},
}, initialState);
```

A wild guess is that it's getting tripped by the `<` and `>` in the generics method declaration, but that seems... silly.

Anyone know what I'm doing wrong? The tsx-ish examples I've found are all over the place and they just mention automatic ts/tsx interpretation.

My tsconfig, for reference:

```json
{
	"compilerOptions": {
		"target": "es6",
		"module": "commonjs",
[...]

PR #216, opened 2017-02-17
Pass the source file name to the transpiler

zeh:
When transpiling files as modules, the filename is passed to TypeScript so it can interpret it properly. This is necessary to respect the behavior of the command line, especially when deciding whether files should be interpreted as jsx/tsx files, as per [the transpiler reference](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#transpiling-a-single-file).

See issue #213 for the original report.


nchanged:
That's just great! Thanks!

PR #218, opened 2017-02-18
Loader Types

basarat:
Users would do: 

```ts
/// <reference path="../node_moduels/fuse-box/dist/LoaderAPI.d.ts" />
```

:rose:


nchanged:
Thanks!

devmondo:
@basarat  personally i am sick of TS script reference, isn't there a better way to do this? i mean with all the improvements Microsoft did to the typing support, like install through npm,etc.. we still have to do this ?

basarat:
> personally I am sick of this TS type reference style, isn't there a better way to do this

I agree 💯  but the issue is in our code base. The issue is that the file injects a global so we can't do `import {FuseBoxLoader} from "somewhere"`

To add *globals* user needs to use either `<reference` or explicitly list the file in our `tsconfig.json` as there is no *js way* to link from user code to point to the *global* declaration file.

Note that the loader API doesn't need to be global (`import {Loader} from "fusebox/loader"` can be made to work), just that it is global right now and changing it will be a breaking change :rose:

devmondo:
@basarat  makes total sense!, can't we just extend the `tsconfig` when we get it from user to dynamically add those types before we start fusing, or maybe i am making crap sense :)

basarat:
https://github.com/fuse-box/fuse-box/pull/232/commits/aa156b5bb4ff49eb6e990daa6a2646e534cc944e :rose:

devmondo:
@basarat  you rock 💯

Issue #207, opened 2017-02-17
Make HMR configurable


wagerfield:
I propose that HMR be made optional/configurable when running the `devServer`, since not everyone is a fan of HMR.

In much the same way that you have a `cache` flag for enabling/disabling FuseBox's cache, `hmr` could enable/disable HMR when using the `devServer`.

As per the discussion on Gitter, FuseBox's cache is always enabled when `hmr: true` (since it is required by HMR).

The config might look like this:

```js
FuseBox.init({
  hmr: false,
  cache: true,
  homeDir: 'src',
  outFile: 'bundle.js'
}).devServer('>index.js')
```

Having cache set to `true` would allow FuseBox to rebundle files rapidly in memory, but not update/inject the changes over the websocket. 

Perhaps a log could still be printed in the console prompting the user that files have been updated and to refresh the page to see the changes (like webpack does)?

Conversely another valid configuration might be as follows:

```js
FuseBox.init({
  hmr: true,
  cache: false, // This is overridden when using the devServer with hmr: true
  homeDir: 'src',
  outFile: 'bundle.js'
}).bundle('>index.js')
```

I'm not sure if there is much/any benefit to the above when simply calling `bundle`? Beyond faster bundle times, is there any other benefits to disabling the `cache`?

nchanged:
Thanks!
1) `hmr` and `bundle` cancel eath other
2) `hmr` false can be introduced. I agree

basarat:
> Perhaps a log could still be printed in the console prompting the user that files have been updated and to refresh the page to see the changes (like webpack does)?

Can be done now using the loader plugin api : http://fuse-box.org/#loader-plugins 

![image](https://cloud.githubusercontent.com/assets/874898/23142397/d67ce068-f80f-11e6-82ff-4d04368184c5.png)

:rose:

nchanged:
@basarat could you fix the typings? Travis does not like something.

nchanged:
:rose:

basarat:
45 mins. Till I get to a computer 🌹 Ref for me https://travis-ci.org/fuse-box/fuse-box/builds/203604036

basarat:
https://github.com/fuse-box/fuse-box/pull/255 :rose:

nchanged:
https://github.com/fuse-box/fuse-box/commit/c3d4cf3f5b29d309aec2f3d23ab10f4a1b0ea68c

PR #240, opened 2017-02-19
Cleanups / docs

basarat:
* Docs to disable HMR (requested here https://github.com/fuse-box/fuse-box/issues/207)
* Rename `check` to `isStateful` as it conveys the meaning if its return value better :rose:


nchanged:
I think you can sync =)

basarat:
> I think you can sync =)

synced :rose: :heart:

PR #246, opened 2017-02-20
move tests to typescript

basarat:
:rose:


basarat:
@nchanged once this is merged I will proceed to cleaning up the test directory further :rose: e.g. 

- we don't need to use the built `commonjs` version of the our src 
- a lot of the `done` callbacks can just be replaced with a promise chain (mocha natively supports failing tests on promise failures : https://mochajs.org/#working-with-promises )

As always thanks in advance :heart:

nchanged:
Cool! How does one run just one mocha file?

basarat:
> Cool! How does one run just one mocha file?

`describe.only` :rose:

basarat:
image

basarat:
@nchanged added support for terminal `npm run test:only -- test/something.ts` :rose:

nchanged:
Tests are coming soon -)

PR #255, opened 2017-02-20
remove return as `void` is not compatible with `boolean`

basarat:
As requested https://github.com/fuse-box/fuse-box/issues/207#issuecomment-281185590 :rose:


nchanged:
oh.. so that was me. Sorry =)

Issue #270, opened 2017-02-21
Plugins incorrectly applied


indrimuska:
### Problem

Using a configuration file that makes use of conditionally applied plugins (plugins that needs to be applied if the given RegExp success) will return an unexpected bundle process.

### Configuration

Given the following directory structure, I have a very simple Angular2 application and I want to bundle all CSS files contained in the `styles` folder in an single stylesheet bundle. On the other side, I want to merge all the stylesheets inside my application, so basically injecting their content in the single javascript bundle.

```
angular2-test-app/
  ├── src/
  │   ├── app/
  │   │   ├── app.component.css
  │   │   ├── app.component.ts
  │   │   └── app.module.ts
  │   ├── styles/
  │   │   └── main.css
  │   └── main.ts
  ├── fuse.js
  └── package.json
```

`main.ts` is responsible of including the `main.css` file in the project:

```javascript
import './styles/main.css';
```

That being said, that's the configuration file to run fuse-box.

```javascript
// fuse.js
const fsbx = require('fuse-box');

fsbx.FuseBox.init({
    cache: false,
    homeDir: 'src/',
    outFile: './dist/app.js',
    plugins: [
        fsbx.TypeScriptHelpers(),
        fsbx.HTMLPlugin({ useDefault: false }),
        [ /app(\/|\\).+\.css$/, fsbx.RawPlugin() ],
        [ /styles(\/|\\).+\.css/, fsbx.CSSPlugin({ group: 'app.css', outFile : "dist/app.css" }) ],
    ]
}).bundle('>main.ts');
```

### Result & Analysis

As I mentioned in #177, after running `node fuse.js` I notice that no CSS file were produced in the `dist` folder. But sometimes (apparently random) it will, so I start debugging this issue.

I was able to recreate two working environments:

 1. [angular2-test-app.zip](https://github.com/fuse-box/fuse-box/files/791384/angular2-test-app.zip): bugged version to reproduce the issue
 2. [angular2-test-app-2.zip](https://github.com/fuse-box/fuse-box/files/791385/angular2-test-app-2.zip): not affected version

These two package are not apparently identical, they are completely identical!
Anyway I started debugging and for the first project the `CSSPluginClass::transformGroup` was never called. Neither `CSSPluginClass::transform` was called! On the other side, project number 2 fires every time those methods. My researches end with this screenshot:

[...]

PR #272, opened 2017-02-21
Fix RegExp test subject for plugins to apply

indrimuska:
Replaced full path in favor of relative path (from `homeDir`).
Ref. #270


nchanged:
I did not get, what is it exactly you have solved here?

indrimuska:
Check out problem and solution in #270

nchanged:
Brilliant!!! Could you be so kind and write a few test cases?

indrimuska:
I don't think of being able to.

nchanged:
No worries man! Thanks a lot!

PR #276, opened 2017-02-22
Allows testing plugin RegEx even if the `homeDir` setting is set

indrimuska:
If `homeDir` is set, it was not possibile to test node_module folder.
We should use app-root-path instead.

## Ref. #270, #272


nchanged:
Nice catch! Thanks

PR #281, opened 2017-02-22
[TSX] Allow TSX files to be renamed back to TSX inside source maps

### LINKED ISSUE #280: Sourcemaps for TSX files lose original name when being passed through babel
We use TypeScript with React + Relay, which means we run our tsx files through babel as well as tsc. We have sourcemaps working through both TypeScript and Babel, and that works 👍 

The only issue is that right now the hand-off process from TS -> Babel changes the file names inside the sourcemaps to refer to non-existent files.

I have an OSS example, https://github.com/artsy/reaction-force/tree/orta-sourcemaps ( specifically  https://github.com/artsy/reaction-force/commit/9f4b4ba8ff8f7ebb6ebc22f375f60b132e02d6b4 ) 

Which you can run via

``` sh
git clone https://github.com/artsy/reaction-force
cd reaction-force
git checkout 9f4b4ba8ff8f7ebb6ebc22f375f60b132e02d6b4
yarn install
node fuse.js
```

It will generate a file `reaction-force.js.map` which looks like this:
```json
{"version":3,"sources":["components/__stories__/artwork.jsx","components/artwork.jsx","relay/config.ts","relay/root_queries.ts","components/__stories__/artwork_grid.jsx","components/artwork_grid.jsx","packages/default"],"names":[],"mappings":";;;;;;;;;;;;;;;;AAAA,+BA...
```

For it work correctly, it would need to look like:
```json
{"version":3,"sources":["components/__stories__/artwork.tsx","components/artwork.tsx","relay/config.ts","relay/root_queries.ts","components/__stories__/artwork_grid.tsx","components/artwork_grid.tsx","packages/default"],"names":[],"mappings":";;;;;;;;;;;;;;;;AAAA,+BA...
```

If I manually make these changes, everything hooks up perfectly

orta:
I couldn't think of a regex that covers both JS & JSX, and supports correctly adding the x or not. 

Fixes #280

### DIFF
--- a/src/File.ts
+++ b/src/File.ts
@@ -346,7 +346,7 @@ export class File {
         if (result.sourceMapText && this.context.sourceMapConfig) {
             let jsonSourceMaps = JSON.parse(result.sourceMapText);
             jsonSourceMaps.file = this.info.fuseBoxPath;
-            jsonSourceMaps.sources = [this.info.fuseBoxPath.replace(/\.js$/, ".ts")];
+            jsonSourceMaps.sources = [this.info.fuseBoxPath.replace(/\.js(x?)$/, ".ts$1")];
             result.outputText = result.outputText.replace("//# sourceMappingURL=module.js.map", "")
             this.sourceMap = JSON.stringify(jsonSourceMaps);
         }


alloy:
This could work:

```
> "foo.jsx".replace(/\.js(x?)$/, ".ts$1")
'foo.tsx'
> "foo.js".replace(/\.js(x?)$/, ".ts$1")
'foo.ts'
```

orta:
Thanks @alloy  - updated

nchanged:
Splendid! thanks!

Issue #278, opened 2017-02-22
Sass Includepaths keeps growing causing import issues


ConneXNL:
https://github.com/fuse-box/fuse-box/blob/master/src/plugins/SassPlugin.ts#L30-L38

The following lines create a situation where options.includePaths === this.options.includePaths which results in the includePath growing which each SCSS compile.

This results in all sorts of errors as the includepaths will contain an array of false paths for each SCSS file after the first.

The issue I had because of this was that "./mixin.scss" resolved to the wrong error thus resulting in "Error: no mixin named xxx" errors.

nchanged:
Thanks!
Could you propose a change?

ConneXNL:
Here you go https://github.com/fuse-box/fuse-box/pull/285

PR #285, opened 2017-02-22
Fixes user-defined sass includePaths from being overwritten

ConneXNL:
Fixes https://github.com/fuse-box/fuse-box/issues/278


nchanged:
Thanks, man, now i need to check what's wrong with tests ;-)

PR #329, opened 2017-03-01
Specify import to do in example

JiDai:


aretecode:
@JiDai https://github.com/fuse-box/fuse-box/blob/master/docs/configuration.md#full-config

Issue #291, opened 2017-02-23
Add support for fluent api style


aretecode:
looking up key vals in http://fuse-box.org/#arithmetic-instructions can be quite tedious, and so I propose a fluent api style to use as an optional alternative.

```js
fuse
.startBundle('coolbundle')
.ignoreDeps()
.add('fuse.magic.ts')
.addMatching('*/**.js')
.include('path')
.include('fs')
.finishBundle()

.startBundle('webworker')
.includeDeps() // default
.exclude('fs')
.add('webworkerfile')
```

PR #336, opened 2017-03-03
[wip] extract part of fluent-api from flipbox

aretecode:


nchanged:
Cool! But but why it's in a text file? )

aretecode:
https://github.com/fuse-box/fuse-box/issues/291

leon-andria:
Can we amend before merged?

aretecode:
@leon-andria @nchanged where should file go? how should it be exposed?

leon-andria:
The fluent API should be independent so it goals is abstract the parsed syntax. 
This means that the api will return at the end the parsed arithmetic string. We need to take benefit of the current implement that is already well tested.
So the bundle should accept in parameters the arithmetic parsed object rather than just a string.
This will keep consistency of the current exposed bundle method.
For the fluent api, it may like below:
BundleExpression.startsWith('entry.ts', withDeps:boolean=false)
.include(['**/*.html'], withDeps:boolean=false)
.exclude(['moment'])
.packAs ('fuse-magic')

Does this make sense and unify the existing implementation?

devmondo:
i think we should have `exclude` and `excludeDeps`

aretecode:
this still doesn't answer the question, where should this file go? how is it exposed? fsbx.BundleExpression?

leon-andria:
I think its good to put it under arithmetic folder under bundlerExpression.ts file. And the final result returns the PropParser object.
Overload methods bundle and initateBundle with typed PropParser type under https://github.com/fuse-box/fuse-box/blob/master/src/core/FuseBox.ts

Issue #355, opened 2017-03-06
Change sourcemaps to sourceMaps


wagerfield:
Babel and TypeScript both use camel-cased `sourceMaps: true|false` in their configs. I propose that FuseBox follow these conventions.

```js
FuseBox.init({
  sourceMaps: true // not sourcemaps :)
})
```

The docs and migration warning should also be updated to reflect this.

aretecode:
yay for compatibility!

PR #357, opened 2017-03-06
Changed sourcemaps to sourceMaps

wagerfield:
Fixes https://github.com/fuse-box/fuse-box/issues/355


wagerfield:
**NOTE:** I haven't updated any docs, tests or code for `SourceMapPlainJsPlugin`. I wasn't entirely sure what to do here, so if this needs updating also—please guide me.

nchanged:
Thanks! We need to keep `sourcemaps` as well, In order for people to migrate. You think you can keep it there?

wagerfield:
Updated `sourceMaps` logic to support `sourcemaps` too as requested.

wagerfield:
Not related to the title of this PR, but there is a fix for the `ReplacePlugin` too. This small update will replace all matches as opposed to just the first occurrence.

src/core/FuseBox.ts
```diff
@@ -177,18 +178,19 @@ export class FuseBox {
             //this.context.sourceMapConfig = opts.sourceMap;
         }
 
-        if (opts.sourcemaps) {
+        const sourceMaps = opts.sourceMaps || opts.sourcemaps
```
nchanged: Great!

src/plugins/ReplacePlugin.ts
```diff
@@ -9,7 +9,8 @@ export class ReplacePluginClass implements Plugin {
     transform(file: File) {
         for (let key in this.options) {
             if (this.options.hasOwnProperty(key)) {
```
wagerfield: Is this condition needed?

nchanged: Typescript ... safe properties =)

PR #368, opened 2017-03-08
start cli + slightly optimize loader api

aretecode:
start cli
fusebox
help

add commander and inquirer dev deps
add help to cli 
add todos for cli

add interactive cli using inquirer
add config choices for interactive cli 
add config steps with nested hierarchical interactions
add utils for cli: command filtering, util inspecting, log instantiation, child process invoking for help commands using current stdio for output
minor comment formatting that could not be autofixed by linter 
adding optional type to Log to allow usage without a full workflow context

loaderapi with hoistable fns (smaller slightly more optimized loader api)
adding more of a symbols index to docs for arithmetics
adding fluent arithmetics to docs


src/Log.ts
```diff
@@ -11,15 +11,21 @@ export class Log {
     private totalSize = 0;
     private printLog = true;
 
-    constructor(public context: WorkFlowContext) {
+    constructor(public context: WorkFlowContext | any) {
```
nchanged: it can't be any

bin/fsbx.js
```diff
@@ -0,0 +1,5 @@
+const path = require("path");
+const root = path.join(__dirname, "../");
+process.env.FUSEBOX_DIST_ROOT = root;
```
nchanged: this can be removed (put it there cuz it is required by fusebox bundle)

src/Log.ts
aretecode: that's about as vague as `any` is, why?

nchanged: No, it can be WorkFlowContext only

bin/fsbx.js
aretecode: https://github.com/fuse-box/fuse-box/pull/368/commits/39765f1dfc5609033c4aeb0e88f9f90cd7e1a52c#diff-ecd92f5709aad339d83c281d5a01477dR9

PR #379, opened 2017-03-09
Simply plugin transform runner, and added support for multiple plugin…

### LINKED ISSUE #378: Only first plugin that matches Plugin.test regex is used
I am tring to use 2 plugins that both work in js files, howver only the first one is ever being applied.

See:
https://github.com/fuse-box/fuse-box/blob/master/src/core/File.ts#L208
and
https://github.com/fuse-box/fuse-box/blob/master/src/core/File.ts#L226

It just finds the first match, once found, stop looking for more plugins, btw that whole section of code seems rather complicated. I will have a go at making a fix for it, but i am a little concerned that this is by design.

cseufert:
Simply plugin transform runner, and added support for multiple plugins on same source file

Should fix #378

### DIFF
--- a/src/core/File.ts
+++ b/src/core/File.ts
@@ -203,47 +203,55 @@ export class File {
      */
     public tryPlugins(_ast?: any) {
         if (this.context.plugins) {
-            let target: Plugin;
-            let index = 0;
-            while (!target && index < this.context.plugins.length) {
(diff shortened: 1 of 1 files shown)


cseufert:
Sorry I missunderstood how plugins were processed.

PR #386, opened 2017-03-11
cli

aretecode:
- add version number
- add cli 
- add test for cli


aretecode:
@nchanged ping

package.json
```diff
@@ -5,22 +5,31 @@
...
         "dist",
         "modules",
         "test",
         "docs"
     ],
     "scripts": {
+        "test-cli": "fusebox",
         "help": "node bin/fsbx --help",
-        "test": "TRAVIS=true ./node_modules/mocha/bin/mocha && node ./test.js",
+        "lint-fix": "eslint src/*.ts src/**/*.ts src/**/**/*.ts test/*.js test/**/*.js --fix",
```
Aladdin-ADD: nitpick. it seems a little duplicate.can we change like this:
```bash
eslint --fix --ext .js,.ts src test
```

aretecode: @Aladdin-ADD the reason it is this way is we have to be explicit about what eslint lints, as the typescript parser for eslint is not perfect. If it was, it would only need to be `eslint --fix` as the file extensions are already in the .eslint.

Aladdin-ADD: got it. can we just config the parser to be [typescript-eslint-parser](https://github.com/eslint/typescript-eslint-parser), as it is ESTree-compatible?

Aladdin-ADD: get your point.I've encountered this problem too. The eslint default parser may not work well with typescript. if you don't want to use a experimental parser, I think a better way is to use [tslint](https://github.com/palantir/tslint).

aretecode: there are a *lot* of eslint plugins that tslint does not have. It is not using default eslint parser, I've already set it up in master https://github.com/fuse-box/fuse-box/blob/master/.eslintrc.js#L2 

just the script was removed because there were some files that were autofixed in /modules/ that were fixed perfectly valid, however, they are included in any build chain so they are not compiled to es3 which causes some (very minor) accessibility issues.

However, modules is no longer linted, after the initial fix, the code base has a uniform style.

aretecode: as a side note, I'll likely redo this cli in the future as I've created a *much* better *much* shorter fluent cli with [flipcli](https://www.npmjs.com/package/flipcli)

Aladdin-ADD: Sorry,I didn't see the parser config.Thanks for clarifying. :satisfied:

PR #423, opened 2017-03-22
sourceMaps shouldn't throw an error if outFile is not set

vtabary:
Today, ```init({sourceMap: {}})``` has been deprecated in favor of ```init({sourceMaps: true})```

But this prevent us to generate source-map when using multiple bundles at once

In order to support multiple bundles, the name of the source-map should be generated only if an outFile is mentionned. So it has to support an init object without outFile.


nchanged:
I would hold off going for these changes, as 1.4.1 will deprecate outFile. Please check branch 1.4.1 (there also examples on GitHub) if you and any ideas, let's work on making 1.4.1 better :)

vtabary:
Ok, I worked on master, that's why. 

So I'll check the 1.4.1

nchanged:
Yeah, 1.4.1 has many breaking changes. check the preview repository (on the list in fuse-box)

PR #448, opened 2017-04-01
little tweak for __generator.js for IE9/10/11

matthiasak:
`self(){ ... }` is throwing errors in IE 9, 10, and 11. This small tweak keeps it ES3 compatible.


nchanged:
Nice! thanks

PR #452, opened 2017-04-05
Fix declaration for noImplicitAny

magnushiie:
Omitting the return type of fn causes compile errors in consuming projects that import fuse-box with noImplicitAny compiler option.


nchanged:
Thanks!

PR #458, opened 2017-04-13
LESSPlugin: change filename path from relative to absolute

kwoon:
This patch fix `@import` resolving inside the less file, if main less file isn't in root directory.

### DIFF
--- a/src/plugins/stylesheet/LESSPlugin.ts
+++ b/src/plugins/stylesheet/LESSPlugin.ts
@@ -43,7 +43,7 @@ export class LESSPluginClass implements Plugin {
             less = require("less");
         }
 
-        options.filename = file.info.fuseBoxPath;
+        options.filename = file.info.absDir;
 
         if ("sourceMapConfig" in context) {
             options.sourceMap = { ...sourceMapDef, ...options.sourceMap || {} };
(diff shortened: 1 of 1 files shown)


src/plugins/stylesheet/LESSPlugin.ts
```diff
@@ -43,7 +43,7 @@ export class LESSPluginClass implements Plugin {
             less = require("less");
         }
 
-        options.filename = file.info.fuseBoxPath;
+        options.filename = file.info.absDir;
```
nchanged: But that's a directory not a filename. `absPath` is better maybe?

nchanged: DId you check how well it works with the sourcemaps?
I think we should make it optional.

kwoon: Year, you are right, it was a typo. Of course `absPath`.

nchanged:
Ok, I will merge it now. Thanks,  Please take a look at the 2.0 branch. The release is soon!
There will be many changes to the API https://github.com/fuse-box/preview-1.4.1/tree/master/examples/react-code-splitting

PR #472, opened 2017-05-01
Fixed build error and add process shim for server bundles

TheOriginalAlex:
The collectBundle function is expecting a `Promise<ModuleCollection>` return, but gets a `Promise<void>`.

And when running a server bundle, environment variables don't get injected since the process shim isn't loaded.  But for browsers, people might wanna do vendor bundles that already includes the shim.


nchanged:
So what are the changes then?

TheOriginalAlex:
Added a prepublish `gulp init` so it runs `gulp init` when I try to install fusebox or run `npm pack`.  And fixed the return type of the `collectBundle` function.

PR #473, opened 2017-05-01
fixed bundle splitting bug

TheOriginalAlex:
Fixed the bug where bundle is undefined for bundle splits

### DIFF
--- a/src/core/BundleSplit.ts
+++ b/src/core/BundleSplit.ts
@@ -42,6 +42,7 @@ export class BundleSplit {
         config.plugins = [].concat(config.plugins || [])
         config.standalone = false;
         const fuse = FuseBox.init(config);
+        new Bundle(name, fuse, producer);
         fuse.context.output.setName(joinFuseBoxPath(this.dest, name))
         let conf = new SplitConfig();
         conf.fuse = fuse;
--- a/src/core/WorkflowContext.ts
+++ b/src/core/WorkflowContext.ts
@@ -465,7 +465,7 @@ export class WorkFlowContext {
     protected writeSourceMaps(result: any) {
         // Writing sourcemaps
         if (this.sourceMapsProject || this.sourceMapsVendor) {
-            this.output.write(`${this.bundle.name}.js.map`, result.sourceMap);
+            this.output.write(`${this.output.filename}.js.map`, result.sourceMap);
         }
     }
     public shouldSplit(file: File): boolean {


src/core/WorkflowContext.ts
```diff
@@ -465,7 +465,7 @@ export class WorkFlowContext {
     protected writeSourceMaps(result: any) {
         // Writing sourcemaps
         if (this.sourceMapsProject || this.sourceMapsVendor) {
-            this.output.write(`${this.bundle.name}.js.map`, result.sourceMap);
+            this.output.write(`${this.output.filename}.js.map`, result.sourceMap);
```
nchanged: That's not correct. It has to be `bundle.name` 
It takes it from `fuse.bundle("vendor")` - `vendor` for example

src/core/BundleSplit.ts
```diff
@@ -42,6 +42,7 @@ export class BundleSplit {
         config.plugins = [].concat(config.plugins || [])
         config.standalone = false;
         const fuse = FuseBox.init(config);
+        new Bundle(name, fuse, producer);
```
nchanged: Why?

src/core/WorkflowContext.ts
TheOriginalAlex: But when you do bundle splitting because of the `dest` option in the split config, the filename might not be the same as the bundle name.

src/core/BundleSplit.ts
TheOriginalAlex: If an instance of bundle isn't created for the split, it'll error out wherever this instance of fuse tries to access bundle props like `bundle.name` for example.

TheOriginalAlex: image

nchanged: Well, there needs to be a different solution. As your approach might solve your own issue, but will break everything else.
https://github.com/fuse-box/fuse-box-examples/tree/master/examples/react-code-splitting

TheOriginalAlex: How so?  According to this, the split bundles end up in the directory specified by `dest` within the split config.  Shouldn't the source map files end up in the same directory as the js files?

image

TheOriginalAlex:
I'm gonna close this since it's fixed in TypeCheck plugin, but it also needs to be fixed for sourceMaps.

Issue #125, opened 2017-02-11
Support for Vue components?


bovisp:
I just started using Fuse Box this morning and I absolutely love it!

I use Vuejs in my projects a lot and can get a basic example working easily enough. Is it possible to use separate .vue files for each of my components and then import them through index.js like so?

```
Vue.component('example', require('./components/Example.vue'));

```

bovisp:
Just a slight clarification. What I am looking for is a way to load single-file Vue components that look like this:

```
<template>
    <div class="container">
        <div class="row">
            <div class="col-md-8 col-md-offset-2">
                <div class="panel panel-default">
                    <div class="panel-heading">Example Component</div>

                    <div class="panel-body">
                        I'm an example component!
                    </div>
                </div>
            </div>
        </div>
    </div>
</template>

<script>
    export default {
        mounted() {
            console.log('Component mounted.')
        }
    }
</script>

<style type="sass">
$color = blue

body
    background-color: $color
</style>
```

The renderer should also pickup the `type` attribute from the style tag and apply the appropriate css processor (in this case `node-sass`).

leon-andria:
Does this help you ? https://github.com/fuse-box/fuse-box-vue-seed.
Precisely with import in https://github.com/fuse-box/fuse-box-vue-seed/blob/master/src/main.ts#L2

nchanged:
@bovisp Welcome to FuseBox!

You can use [wildcard imports](http://fuse-box.org/#wildcard-import) and register them from a folder. 
This is what you will get:
```
{ "batch/a.js" : {}, "batch/b.js" : {} }
```

It might help you making it more magical.
On top of that, we will make an official Vue compiler plugin soon.

bovisp:
@nchanged. Thanks for the response. Yeah. Will use wildcard imports for now and will wait on the vue compiler.

Love the fuse-box speed. Webpack  runs a little slow on my 10 year old MacBook Pro, but fusebox just hums along. Keep up the great work.

Paul

smitt04:
@nchanged is there an issue or project to watch for vue.js compiler? I am waiting for that before i can try out fuse-box.

nchanged:
We have a primitive version of it, let us test a bit. You can watch this issue.

rottmann:
Is there a short example on how to use VuePlugin?

i tried:

bundle
```
let app = fuse.bundle('app')
  .instructions(`> index.ts`)
  .plugin(VuePlugin())
[...]

PR #474, opened 2017-05-02
Fix VuePlugin

Sirikon:
As said in Issue #125 on [this comment](https://github.com/fuse-box/fuse-box/issues/125#issuecomment-298455538)

### DIFF
--- a/src/plugins/VuePlugin.ts
+++ b/src/plugins/VuePlugin.ts
@@ -48,12 +48,11 @@ export class VuePluginClass implements Plugin {
             const ts = require("typescript");
 
             const jsTranspiled = ts.transpileModule(jsContent, file.context.getTypeScriptConfig());
-            const tsResult = `var View = require('vue/dist/vue.js');
-var _p = {};
+            const tsResult = `var _p = {};
 var _v = function(exports){${jsTranspiled.outputText}};
 _p.template = ${JSON.stringify(html)};
 var _e = {}; _v(_e); _p = Object.assign(_e.default, _p)
-module.exports ={render : function(el){_p.el = el; return new View(_p)}}
+module.exports =_p
             `;
             file.contents = tsResult;
             file.analysis.parseUsingAcorn();


nchanged:
I kind of didn't get how would one render a Vue component? ;-)

Sirikon:
Hm.. sorry? I don't understand your question... do you mean why does it work using just _p ?

nchanged:
I mean, you have removed the `View` instance. What does it export then? just an object? Can you give an example on how would you use the plugin?

Sirikon:
When using Vue components all you need to give to Vue to render the component is a plain javascript object with, at least, a render method or template property.

So, in this case with this two files:

```js
// index.ts
import * as Vue from 'vue/dist/vue.js'
import * as App from './components/App.vue'

new Vue({
    el: '#app',
    render: h => h(App)
})
```

```html
<!-- components/App.vue -->
<template>
    <div>
        <p>{{ msg }}</p>
        <input type="text" v-model="msg" />
    </div>
</template>

<script>

    export default {
        name: 'app',
        data () {
            return {
                msg: 'Welcome to Your Vue.js App'
            }
        }
    }

</script>
```

This will work because index.ts creates a new Vue instance, which will use the `h` method inside `render` method, here the `h` method is `createElement` from Vue, which can take just a plain javascript object.

Sirikon:
@nchanged Just made a Github repo with a project that will work once this PR is merged: https://github.com/Sirikon/fuse-box-vue-example

nchanged:
Right, I see, so you wouldn't want to export the entire view? Instead just a template?

Sirikon:
@nchanged Check the `<script>` tag content in `components/App.vue`, it just exports that, a plain javascript object. If the developer using this wants to export an actual Vue instance, can just export it instead of the plain JS object.

This is also, if I'm not wrong, the default behaviour in the official tools for webpack.

nchanged:
@Sirikon ok, it's just I know so little about Vue. You think you could write some docs on it?

Sirikon:
@nchanged Sure! :) will write about it and make another PR

nchanged:
Thanks =)

PR #479, opened 2017-05-03
Add VuePlugin documentation

Sirikon:
@nchanged Would be nice to have a repository behind fuse-box organization for a Vue example, just like the Angular2 and React ones, so we can add it as a link in the documentation.


nchanged:
Great job, lets create a seed later!

PR #485, opened 2017-05-05
Allow the use of smaller vue-runtime by compiling templates

CantGetRight82:
https://vuejs.org/v2/guide/installation.html
(see  section "Runtime + Compiler vs. Runtime-only)

By compiling to a render function instead of a template string, users are able to use the smaller runtime version of vuejs.

Largely based on the template compiler of vuejs/vueify.

### DIFF
--- a/src/plugins/VuePlugin.ts
+++ b/src/plugins/VuePlugin.ts
@@ -7,6 +7,7 @@ export interface VuePluginOptions {
 }
 
 let vueCompiler;
+let vueTranspiler;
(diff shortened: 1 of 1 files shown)


nchanged:
Awesome, do you think you could write some tests?
For example https://github.com/fuse-box/fuse-box/blob/master/src/tests/plugins/CSSPlugin.test.ts

And to run it:

```
node test --file=CSSPlugin.test.ts
```

That would help a lot!

PR #487, opened 2017-05-05
Add VuePlugin test

CantGetRight82:


nchanged:
Can we test the actual render? if we refactor it later?

CantGetRight82:
Test with selenium? Or test server side render? I agree that the current test is very exact, but at least there is a test. I think I will be able to put the time in when FuseBox 2.0 is released.

Can you explain how to use the dependencies array in the plugin? I tried to use it but it did not seem to work. As a workaround I have manually added two 'vue-template' modules to the travis yml, but the end user is now still required to manually install the required modules. (as i was when first testing the VuePlugin)

nchanged:
No, we are using JSDOM, so technically you will have access to `document`
And, for example: https://github.com/fuse-box/fuse-box/blob/master/src/tests/plugins/HTMLPlugin.test.ts#L49

You import a file ( the scope is strictly limited) and get the results of your plugin transformation (at runtime)

And `project` is actually [window](https://github.com/fuse-box/fuse-box/blob/master/src/tests/stubs/TestEnvironment.ts#L144)

CantGetRight82:
Thanks for the feedback. I think to be able to test the generated html, I would have to include the vue-runtime in the testing process:

**Old situation**
.vue file  -> backend generates template string -> frontend vue interprets template -> frontend vue renders

**New situation**
.vue file -> backend interprets template -> frontend vue renders

So in the new situation, actual HTML is only generated in the Vue instance.

Is it ok to include vue in the gulpfile.js for testing the actual HTML output?

Thanks for your patience!

nchanged:
@CantGetRight82 You mean add it to devDeps in gulp task? Yeah that's okay ;-)

src/tests/plugins/VuePlugin.test.ts
```diff
@@ -0,0 +1,61 @@
...
+                instructions: ">app.vue",
+            },
+        }).then((result) => {
+			const out = result.project.FuseBox.import("./app.vue").trim();
+			should(out).findString('_p.render');
+			should(out).findString('_p.staticRenderFns');
+			should(out).notFindString('_p.template');
+
+			should(out).equal(`var _p = {};
+var _v = function(exports){"use strict";
```
aretecode: 🔬👍 good to add tests

this type of string output exactly matching unit test for something that may be a frequently updated dependency seems kind of fragile.

why not: 

- compile the content with fusebox (as it currently does) 
- write to disk & require, 
- or use the [node vm sandbox](https://nodejs.org/api/vm.html), 
- then [use vue unit testing](https://vuejs.org/v2/guide/unit-testing.html) to assert the rendered text? 

easier to read, update, and it asserts more things.

CantGetRight82: I couldn't agree more. We were already discussing a better way. Let me see if I can improve it today

PR #494, opened 2017-05-06
Add options to watch() and src() with 'base' param

### LINKED ISSUE #493: Sparky.watch() should have a way to define a basepath
I have a test project using FuseBox with a folder called `src` on root. Inside that, `assets/logo.png`

With the next sparky task:
```js
Sparky.task('copyAssets', () => {
    return Sparky.watch('./src/assets')
        .dest('./dist')
})
```
Will work as expected, create `src/assets/logo` inside `dist`, but I need to copy just `assets` inside `dist`. not `src`.

There's no documented way to to this, but found out that the watch() method can accept a second parameter which is a chokidar watch() method options parameter. So, using this:

```js
Sparky.task('copyAssets', () => {
    return Sparky.watch('./assets', { cwd: './src' })
        .dest('./dist')
})
```

Creates the file structure correcty... except the `logo.png`. Using the debugger, saw that the cwd param is used by chokidar, but isn't used to correctly build the paths that chokidar is returning.

So:

- Sparky watches `assets` inside `src`
- Chokidar watches `assets` inside `src`
- Chokidar returns `assets/logo.png`
- Sparky injects to the flow `assets/logo.png`, instead of `src/assets/logo.png`

Sirikon:
Created a new interface called `SparkyFilePatternOptions` which is like this:

```typescript
export interface SparkyFilePatternOptions {
    base?: string; // Will be ignored when parsing absolute paths
}
```
Added the interface as optional params to src() and watch() methods, and is actually used in parse method inside SparkyFilePattern.ts.

Now something like this would work:
```js
Sparky.task('copyAssets', () => {
    return Sparky.watch('./assets', { base: './src' })
        .dest('./dist')
})
```
And should fix #493

### DIFF
--- a/src/sparky/SparkFlow.ts
+++ b/src/sparky/SparkFlow.ts
@@ -4,8 +4,12 @@ import * as fs from "fs-extra";
 import { ensureDir, string2RegExp } from "../Utils";
 import { SparkyFile } from "./SparkyFile";
 import { log } from "./Sparky";
-import { parse } from "./SparkyFilePattern";
+import { parse, SparkyFilePatternOptions } from "./SparkyFilePattern";
 import * as  chokidar from "chokidar";
+
+export interface SparkFlowWatchOptions {
+    base? : string;
+}
 
 export class SparkFlow {
     private activities = [];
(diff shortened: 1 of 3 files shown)


nchanged:
good idea, but please check the tests, all sparky tests have failed ;-(

Sirikon:
@nchanged Fixed :) There was a typo...

nchanged:
Awesome, thanks!

nchanged:
`fuse-box@2.0.0-beta.14`

devmondo:
I tested it and it works, but we need to update the docs to reflect the new change.

great work @Sirikon!

Sirikon:
Thanks! :)
@devmondo Was going to do that as soon as I get my repo synced

PR #495, opened 2017-05-06
Update VuePlugin and Sparky docs

Sirikon:
Updated Sparky docs to add `options.base` #494 
Updated VuePlugin docs to include the new required module `vue-template-es2015-compiler` #485


nchanged:
Good job! thanks a lot!

PR #496, opened 2017-05-06
Update Sparky tests and fix a test for Windows

Sirikon:
Updated Sparky tests adding a test for `base` option.

Fixed an edge case in Windows in `SparkyFile.ts` constructor, it just needed to `path.normalize()` the root path. This made the test fail in Windows.

Also:
```js
let hp = this.filepath.split(root)[1]; // Removed this
let hp = path.relative(this.root, this.filepath); // And did this instead
```
Which does the intended use and is a safer way to work with paths.


nchanged:
Awesome!

Issue #492, opened 2017-05-06
Support Vue template preprocessors (e.g. Pug)


NGPixel:
It would be great if custom template languages, such as Pug, could be used when using .vue components.

```
# Standard HTML template (works)
<template>
  <div class="panel">{{ msg }}</div>
</template>

# Pug template (doesn't work)
<template lang="pug">
  .panel {{ msg }}
</template>
```

vue-loader handles the `lang` parameter on `<template>` and `<style>` blocks. Fuse-box should provides similar support.

Thanks,
Nick

nchanged:
hi @NGPixel the VuePlugin isn't working 100% as it supposed it, The original plugin is tightly coupled with WebPack, so we are missing some features. Please feel free to improve it.

nchanged:
If you interested in helping - that's the [place](https://github.com/fuse-box/fuse-box/blob/master/src/plugins/VuePlugin.ts#L47) where we could abstract things, and attach different HTML compilers

NGPixel:
Sure, I'll implement it.

nchanged:
Closing this one

PR #497, opened 2017-05-06
Added .vue components template preprocessing

NGPixel:
As discussed in #492, adds custom template engine to be used by defining the `lang` attribute on the `<template>`  tag.

Also updated the docs for VuePlugin.

### DIFF
--- a/src/plugins/VuePlugin.ts
+++ b/src/plugins/VuePlugin.ts
@@ -81,11 +81,11 @@ function toFunction (code) {
 }
 
 function compileTemplateContent (context: any, engine: string, content: string) {
-    const cons = require('consolidate');
     return new Promise((resolve, reject) => {
-        if (!engine || !cons[engine]) {
-            return content;
-        }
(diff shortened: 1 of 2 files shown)


src/plugins/VuePlugin.ts
```diff
@@ -74,6 +80,23 @@ function toFunction (code) {
   return vueTranspiler('function render () {' + code + '}')
 }
 
+function compileTemplateContent (context: any, engine: string, content: string) {
+    const cons = require('consolidate');
```
nchanged: Shouldn't this be installed as well? Maybe we could mention it in the docs as well

nchanged: I mean we should not add it to the package.json it has to be optional.

devmondo: fully agree

NGPixel:
You're completely right. I added the missing info in the docs.

PR #499, opened 2017-05-07
Fix VuePlugin

Sirikon:
Right now, the published version in NPM (2.0 beta 15 atm) breaks when the Vue template doesn't use any templating engine.

This fixes that, adding the missing `resolve()` call


nchanged:
Awesome! thanks!

PR #502, opened 2017-05-08
Fix compatibility with uglify-js 3

CantGetRight82:
This fix enables the use of either uglify-js 2 or uglify-js 3


nchanged:
thanks a million!!

PR #503, opened 2017-05-08
Now src() and watch() both accept an array

Sirikon:
Also added a test for usage of src() with an array.


nchanged:
Amazing! Thanks!

Issue #513, opened 2017-05-10
SourceMap files path


eddow:
When making bundles in subdirectories, the sourceMappingURL gets the path based on the "dist" root folder

example:
with `fuse.bundle("server/express")`,the last line of `dist/server/express.js` is:
`//# sourceMappingURL=server/express.js.map`

I am using the module `source-map-loader` as it seems nothing was planned natively in node nor fuse-box to use the sourceMap server-side, but therefore that relative path is problematic : the relative path to the map is `express.js.map` instead of `server/express.js.map`

Did I miss something about using the sourceMap on the server-side?

devmondo:
Thank you very much :)

PR #514, opened 2017-05-10
SourceMap path relativity

eddow:
https://github.com/fuse-box/fuse-box/issues/513

Sourcemap path should be relative to source-file path, not to root distribution path.
Note: The issue specifies the problem on the server-side, while it is problematic also on the client-side

### DIFF
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -226,7 +226,7 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
         }
         if (this.context.sourceMapsProject || this.context.sourceMapsVendor) {
             let sourceName = /[^\/]*$/.exec(this.context.bundle.name)[0];
-            this.concat.add(null, `//# sourceMappingURL=${sourceName	}.js.map`);
+            this.concat.add(null, `//# sourceMappingURL=${sourceName}.js.map`);
         }
     }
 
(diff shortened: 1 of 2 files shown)


eddow:
Another problem solved :
When using Vue modules AND generating sourcemap,
The vue source ended with `//# sourceMappingURL ...` , so `{${jsTranspiled.outputText}};` would "lose" its tail  `};`

src/BundleSource.ts
```diff
@@ -225,7 +225,8 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
             this.concat.add(null, "(FuseBox)");
         }
         if (this.context.sourceMapsProject || this.context.sourceMapsVendor) {
-            this.concat.add(null, `//# sourceMappingURL=${this.context.bundle.name}.js.map`);
+            let sourceName = /[^\/]*$/.exec(this.context.bundle.name)[0];
```
nchanged: Should bundle name never start with slash?! Are you sure about the fix?

eddow: As I described in the issue, we need to keep the file name out of the bundle name :
`/[^\/]*$/.exec('path/to/bundle')[0]  ==>  'bundle'

nchanged: But that isn't correct. Correct me if am wrong but, `this.context.bundle.name` === `vendor` or `app`

eddow: In basic cases yes.
In this example already, you will have `server/bundle` and `client/app`
https://github.com/fuse-box/fuse-box-examples/blob/master/examples/server-client/fuse.js

nchanged: Right I see! That's correct. Did you test it yourself? How about the actual source maps file? is it all consistent?

PR #521, opened 2017-05-11
proc.require()

eddow:
Alternative to .completed(proc=> proc.start()) that runs the script in the node instance used by fuse (debugging if fuse is debugged)
```
.completed(proc=> {
	proc.require().then(exps=> {
		// exps is the main module' exports
	});
});
```
There are several ways to express how to close the deprecated instance once the bundle has been re-transpiled, when it is started again, depending on the AEFL (Automatically Executed File on Load)
If the bundle has NO AEFL, a close function taking no argument can be given as argument of `require`
```
proc.require(function() {
	console.log('unloaded');
});
```
If the bundle HAS an AEFL, the first of these choice that exists/is defined will be used :

A- Give a close function to the `require` call, who takes an argument containing the exports of the AEFL
```
proc.require(exps => exps.kill())
```

B- export a close function in the AEFL that takes no argument

```
// server.js
[ ... ]
export function close() {
	myResources.$destroy();
}
[ ... ]
```

C- Export directly in exports or in default an object that hase a close function (without parameters)

```
// server.js
[ ... ]
export default app.listen(config.port);
[ ... ]
```
OR

```
// server.js
[ ... ]
export = app.listen(config.port);
[ ... ]
```

PROMISES:
If the `close` function that is called returns a Promise, the module will be re-started on that promise' resolution.
```
proc.require(function() {
	return Promise((resolve, reject)=> setTimeout(()=> resolve(true), 60000));
});
```
This will wait a minute before re-requireing the module.
note: This promise' resolution value is ignored
OBVIOUS NOTE:
The `close` function is not called the first time the module is launched. It is called just after re-transpilation and before re-launching the module.
RETURN VALUE:
The function `require` returns a Promise that resolves to the exports of the AEFL

### DIFF
--- a/src/FuseProcess.ts
+++ b/src/FuseProcess.ts
@@ -52,7 +52,12 @@ export class FuseProcess {
 				if(!(closePromise instanceof Promise))
 					closePromise = Promise.resolve(true);
 				closePromise.then(
-					()=> resolve(getMainExport(require(this.filePath))),
+					() => {
+						var exps = false;
+						try { exps = getMainExport(require(this.filePath)); }
+						catch(x) { reject(x); }
+						if(exps) resolve(exps);
+					},
 					x=> reject(x)
 				);
 			});
--- a/src/FuseProcess.ts
+++ b/src/FuseProcess.ts
@@ -41,7 +41,7 @@ export class FuseProcess {
 									cached.default.close() : //if a `close` function is exported by the default export
 								console.warn(`Bundle ${this.bundle.name} doesn't export a close() function and no close was given`);
 						} catch(x) {
-							console.error(`Exception while closeping bundle ${this.bundle.name}.`);
+							console.error(`Exception while closing bundle ${this.bundle.name}.`);
 							reject(x);
 						}
 					} else if(close) {
--- a/src/FuseProcess.ts
+++ b/src/FuseProcess.ts
@@ -1,5 +1,6 @@
 import { Bundle } from "./core/Bundle";
 import { ChildProcess, spawn } from "child_process";
+declare function require(name:string);
 
 export class FuseProcess {
     public node: ChildProcess;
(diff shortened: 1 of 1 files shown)


src/FuseProcess.ts
```diff
@@ -22,6 +23,40 @@ export class FuseProcess {
         return this;
     }
 
+		require(close?: ((exports: any) => void)|(() => void)) {
+			function getMainExport(mdl) {
+				return mdl && mdl.FuseBox && mdl.FuseBox.mainFile ?
+						mdl.FuseBox.import(mdl.FuseBox.mainFile) :
```
nchanged: I don't think we need to import it manually. Once a bundle is loaded it's gets executed right away, unless you want to get the results?!

nchanged:
Awesome!
We need a documentation for this feature with examples, you think you could do that?

src/FuseProcess.ts
eddow: Well, the exports of the module are used in 2 cases :
1- as the value of resolution. This one is optional but the exported values are given to the script in fuse - this could be useful for someone, somebody who write a `close` function in the fuse file for example
2- to find back the `close` functions indeed, in order for instance to stop express to be able to re-launch it afterward

eddow:
For the documentation, I did it mainly in the description I gave here - now I could perhaps write in your documentation directly.
For examples - I would need ideas or some bone to bite, perhaps something in https://github.com/fuse-box/fuse-box-examples ?

nchanged:
@eddow yeah, that repo. I will dwell on the code a little bit longer and will merge soon enough ;-)

eddow:
Don't rush - I come again on little modifications I could bring indeed, I don't know if I can still modify it after you pull it

nchanged:
@eddow just lemme know where you are ready. But it's looking cool, I like it. But we need a solid documentation on that which answers the questions `why` `motivation` and `how` ;-)

src/FuseProcess.ts
eddow: Indeed, I just understand your question : the exports are supposed to be exposed in "mdl" - that's because of a bug I think, I'll try to make a repository to expose the bug before we pull this

eddow:
I'll wait for #522 as I think it is the reason I had to hack with `getMainExport`.
For the doc, I just get you a md file I guess? I'm noob in these too but I can give it a shot even if I feel I exposed the main technical points in the description I made here

nchanged:
@eddow that will be this file 
https://github.com/fuse-box/fuse-box/blob/master/docs/bundle.md#executing-a-bundle

 =)

src/FuseProcess.ts
nchanged: http://fuse-box.org/page/configuration#global-variables

nchanged: I am not saying your PR is wrong, it's great - for development ;-)

eddow:
I was going to finish the doc and wait for #522 to redo the `getMainExport` stuff - so if you happen to change the branch by the time, keep me informed

PR #527, opened 2017-05-13
proc.require() doc

eddow:
Sorry, I wrote some doc only afterward.

I don't know why Git says  src/FuseProcess.ts was modified again - if needed, just c/p the help-text and "refuse" this PR.

Please, do not hesitate to rephrase, reformat, etc.


nchanged:
Cool, thanks i will merge it soon, after editing a bit =)

eddow:
Indeed, after I changed as we need the module's FuseBox export to access the "files" from the bundle. I would still wait for the resolution of the export stuff before finalizing the arguments - I have the impression it would change a lot.

Here, I changed also the callback given to `require` for it not to be a function but an options object (with the callback) to let expansions possible

nchanged:
There are some conflicts.

eddow:
Replaced by https://github.com/fuse-box/fuse-box/pull/546

PR #528, opened 2017-05-14
fix `yarn add` command

BrianSipple:
The syntax for adding a package with `yarn` and saving it to `devDependencies` is `yarn add <name> --dev`.

It's a bit tricky to catch, though, because `yarn add <name> --save-dev` still works, and doesn't throw any warnings, but then actually saves the package to `dependencies` 😺.


nchanged:
Thanks :)

PR #539, opened 2017-05-16
Deps grabber

eddow:
Please, be harsh on the review, I really must have missed many test-cases

The plugin must be created with an argument that is :
- a "basket" = an Object that will receive the list of dependencies per bundle
- a callback that will be called for all dependencies

A possible problem is that I don't check if the plugin is called before or after the file analysis
Another is that I didn't test with .js file


eddow:
Hmmmm, could anyone have a look at the test that fails to tell me what is wrong with the plugin ? Or should I make a 3rd-party plugin?

eddow:
I come back to you this evening - let's plan a little chat - my `amd` idea was a dead-end and I found something much simpler that will solve this "dependency grabbing" too

devmondo:
good work @eddow  :)

PR #546, opened 2017-05-18
Proc require

eddow:
Replacement of https://github.com/fuse-box/fuse-box/pull/527 as I had to re-arrange the branches here

Basically - proc.require got his args stabilised and its doc.


nchanged:
Better, now we need to test it =)

Issue #518, opened 2017-05-11
XHR and error reports


eddow:
When playing around or trying to configure heuristically, fuse-box could have more barriers to help you find a way.
There is one about the source-file loading : some source files are just present in the sourcemap after bundling but are unavailable for download.
I happen on some typos to have the browser part of Fuse downloading infinitely each 1/2s the same file (that doesn't exist)
The infinite loop is quite short and could be short-circuited and have a fool-check to stop this behaviour and display an error message
I c/p here the stack trace of the prettyfied uglyfied javascript .. it's few but it's what we have.

```
    function s(e, r) {
        if (!d)
            return r(/\.(js|json)$/.test(e) ? p.require(e) : "");
        var n = new XMLHttpRequest;
        n.onreadystatechange = function() {
            if (4 == n.readyState)
                if (200 == n.status) {
                    var i = n.getResponseHeader("Content-Type")
                      , o = n.responseText;
                    /json/.test(i) ? o = "module.exports = " + o : /javascript/.test(i) || (o = "module.exports = " + JSON.stringify(o));
                    var a = t("./", e);
                    m.dynamic(a, o),
                    r(m.import(e, {}))
                } else
                    console.error(e, "not found on request"),
                    r(void 0)
        }
        ,
        n.open("GET", e, !0),
        n.send()
    }
```
on receive, `m.import(e, {})` leads here :
```
        r.import = function(e, r) {
            return c(e, r)
        }
```
Then :
```
    function c(e, r) {
        [ ... ]
        if (!i) {
            var m = "function" == typeof r
              , x = l("async", [e, r]);
            if (x === !1)
                return;
            return s(e, function(e) {
                return m ? r(e) : null
            })
        }
        [ ... ]
    }
```

Where the call `s(e, function(e) {` closes the loop and begins a new XHR request

nchanged:
That's a tricky one. I am sure your request returns `200OK` that's why fusebox is confused. Try setting it to `404`

eddow:
Indeed, the infinite loop is solved in my case (and no 404 allowed in SPA, it returns the main index.html on unknown route)
This is not about how I should do not to get in an infinite loop, these "barriers" are comfort for the ones that are still uneasy with fuse and make mistakes without knowing : a simple error message can save them 1/2h of searching what might be wrong in their config (I did spend time wondering where this ame from ^^ )
[...]

PR #549, opened 2017-05-19
Fix infinite requests caused by server's 200OK

eddow:
Solving: https://github.com/fuse-box/fuse-box/issues/518

When a module is not found or badly loaded in the browser, if there were no server errors (status: 200), FuseBox will use the response (and fail to register the module badly given) then import the module again ... as the module was not properly registered, the import will lead to an ajax request, and this will repeat indefinitely.

The correction brought here is simple (the infinite recursion is short) : The import will pass the option `ajaxed` to specify the file that was required to lead to that import. The function `$async` will, before creating an XHR object, check if by chance the request is not the one just done - and if yes, signal an error instead of repeating it.

### DIFF
--- a/modules/fuse-box-loader-api/fusebox.js
+++ b/modules/fuse-box-loader-api/fusebox.js
@@ -174,8 +174,12 @@ function $getRef(name, o) {
     };
 }
 ;
-function $async(file, cb) {
+function $async(file, cb, o) {
+    if (o === void 0) { o = {}; }
     if ($isBrowser) {
+        if (o && o.ajaxed === file) {
+            return console.error(file, 'does not provide a module');
+        }
         var xmlhttp = new XMLHttpRequest();
         xmlhttp.onreadystatechange = function () {
             if (xmlhttp.readyState == 4) {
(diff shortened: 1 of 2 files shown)


devmondo:
lovely work @eddow !!!

nchanged:
Good job @eddow !

nchanged:
@eddow so it is ready? or shall I hold off merging for now?

eddow:
Indeed, it is ready :
- I didn't notice any other infinite ajax calls indeed - I did when I removed my modification by mistake
- if you don't long for tests, I'll skip this part and it's well finished

modules/fuse-box-loader-api/fusebox.js
```diff
@@ -192,7 +196,7 @@ function $async(file, cb) {
                     }
                     var normalized = $pathJoin("./", file);
                     FuseBox.dynamic(normalized, content);
-                    cb(FuseBox.import(file, {}));
+                    cb(FuseBox.import(file, { ajaxed: file }));
```
nchanged: That one I like!

PR #557, opened 2017-05-21
Amd modules (package lazy-load)

eddow:
- [x] Test in an actual project
- [ ] Allow lazy auto-exec files (arithmetic `> file.ts`)
- [ ] Documentation
- [ ] Make a fuse-box-example
- [ ] Check for circular references ?

AMD modules allows transparent package lazy-loading.

### Recipe ###
- in `tsconfig.json`, choose `"module": "amd"` instead of `"module": "commonjs"`
- When calling import, give the option `lazyLoadPackage`. Example:
```typescript
import * as load from 'load-script'

declare var FuseBox: any;

function lazyLoadPackage(name: string): Promise {
	return new Promise((resolve, reject)=> {
		load('/myPackages/'+name, (err, script) => {
			if(err) reject(err);
			else resolve(script);
		});
	});
}

FuseBox.import('test-package', {lazyLoadPackage}).then((x)=> {
	console.log('`test-package` exports :')
	console.dir(x);
});
```
### Example ###
Having `test-package.ts` as following: 
```typescript
export {blah} from 'test-import';
```
With the loader from the recipe, will lead to this behaviour :
1. on `FuseBox.import`, `lazyLoadPackage` will be called for `test-package`
2. on reception of `test-package`, it won't be executed directly: first, `lazyLoadPackage` will be called for `test-import`
3. When `test-import` is loaded (considering it doesn't need any other lazy imports) it will be executed
4. `test-import` exports are given to `test-package` that is executed
5. The import promise is resolve and `test-package` exports are written to the console
### Backward compatibility ###
* If a package/bundle had been transpiled before AMD support and is called/used by a bundle using the new version, the local parameters containing the names will have wrong values.

This comes from the arguments change : the amd-less version `(exports, require, module, __filename, __dirname)` with the arguments `(exports, require, module, define, __filename, __dirname)`
* If `lazyLoadPackage` is given, `FuseBox.import` return an export object OR a `Promise`.
  - To make sure it is a `Promise`, one can just write `var importPromise = Promise.resolve(FuseBox.import(...));`
  - If no `lazyLoadPackage` is given, `FuseBox.import` will return an export value or throw an error like before.
* If the `tsconfig.json` doesn't explicitely specify the `amd` module, `commonjs` is kept by default - it has no real advantages but make sure compiled code doesn't change abruptly for users.

Morality : beside for `__filename` and `__dirname`, the PR is completely backward-compatible

### Applications ###
* It works client-side AND server-side - it can be for example combined with [fuse-npm](https://github.com/eddow/fuse-npm) easily to make a completely lazy-loaded source-code runner.
* Core redundancy-proof : `lazyLoadPackage` will never be called twice for the same package name


eddow:
Ok, answered - the problem I was facing was not from windows. Going on...

eddow:
- [x] Test in an actual project
- [ ] Allow lazy auto-exec files (arithmetic `> file.ts`)
- [ ] Documentation
- [ ] Make a fuse-box-example
- [ ] Check for circular references ?

AMD modules allows transparent package lazy-loading.

### Recipe ###
- in `tsconfig.json`, choose `"module": "amd"` instead of `"module": "commonjs"`
- When calling import, give the option `lazyLoadPackage`. Example:
```typescript
import * as load from 'load-script'

declare var FuseBox: any;

function lazyLoadPackage(name: string): Promise {
	return new Promise((resolve, reject)=> {
		load('/myPackages/'+name, (err, script) => {
			if(err) reject(err);
			else resolve(script);
		});
	});
}

FuseBox.import('test-package', {lazyLoadPackage}).then((x)=> {
	console.log('`test-package` exports :')
	console.dir(x);
});
```
### Example ###
Having `test-package.ts` as following: 
```typescript
export {blah} from 'test-import';
```
With the loader from the recipe, will lead to this behaviour :
1. on `FuseBox.import`, `lazyLoadPackage` will be called for `test-package`
2. on reception of `test-package`, it won't be executed directly: first, `lazyLoadPackage` will be called for `test-import`
3. When `test-import` is loaded (considering it doesn't need any other lazy imports) it will be executed
4. `test-import` exports are given to `test-package` that is executed
5. The import promise is resolve and `test-package` exports are written to the console
### Backward compatibility ###
* If a package/bundle had been transpiled before AMD support and is called/used by a bundle using the new version, the local parameters containing the names will have wrong values.

This comes from the arguments change : the amd-less version `(exports, require, module, __filename, __dirname)` with the arguments `(exports, require, module, define, __filename, __dirname)`
* If `lazyLoadPackage` is given, `FuseBox.import` return an export object OR a `Promise`.
  - To make sure it is a `Promise`, one can just write `var importPromise = Promise.resolve(FuseBox.import(...));`
  - If no `lazyLoadPackage` is given, `FuseBox.import` will return an export value or throw an error like before.
* If the `tsconfig.json` doesn't explicitely specify the `amd` module, `commonjs` is kept by default - it has no real advantages but make sure compiled code doesn't change abruptly for users.

Morality : beside for `__filename` and `__dirname`, the PR is completely backward-compatible

### Applications ###
* It works client-side AND server-side - it can be for example combined with [fuse-npm](https://github.com/eddow/fuse-npm) easily to make a completely lazy-loaded source-code runner.
* Core redundancy-proof : `lazyLoadPackage` will never be called twice for the same package name

nchanged:
@eddow I appreciate the effort but you broke many fundamental parts and concepts of FuseBox
* FuseBox.import will never return a Promise
* The API should not be larger than 4.5kb
* Define should NOT be in the core (as most of the people don't use AMD nowadays)
And forcing it's way for everyone is not a solution. An **independent** plugin is.

* bin.js shouldn't be touched by anyone. That's a stable version of FuseBox, created for testing current versions.

Please, if you are planning to introduce something like that - talk to use first on gitter and discuss details. FuseBox is used by many people, and we have to cover many many cases and requirements.

Thanks for understanding.

eddow:
Okay; to be fair, I mainly work for myself here, I propose if I can help : I really need package lazy-loading and `amd` is the straightest way to reach it.
To take back your points :
- As I said, I really need lazyloading of packages, and `Promises` are the straightest way
- I didn't know, indeed, I don't have that constraint for myself ;-)
- I don't know how to bring this amd-style call-back execution of a file without bringing `define` to a file-loader. And I don't know how to bring define to a file-loader with just a plugin. Note: I didn't try, there is perhaps a way indeed.
- Well, I was annoyed to modify generated code but as I was changing the signature of the module function for all the reasons cited here, I didn't have a choice.

It won't be possible to make a plugin out of this, so I will keep (use and maintain) a fork - of course, I give you my advances as PR if you want to integrate them, I never intend to take away your work nor make concurrence to you.

nchanged:
@eddow yes it's possible to make a plugin. You can use [autoImport](http://fuse-box.org/page/configuration#auto-import) to inject something like that:
```
var define = require("fuse-box-amd")
```

I can assist you in getting things done with the approach above ;-)

nchanged:
Regarding package lazy loading, you can check [this](https://github.com/fuse-box/fuse-tools) repository
I think that functionality belongs there

eddow:
Well, the problem is about the transparent lazy-loading. The purpose is for the one who writes modules not to have to worry about how his modules will be loaded.

Transpiling in `commonjs` will transpile imports into a `var ... = require(...);` statement. This statement cannot be "paused" to lazy-load what is required then resumed with the newly loaded requirement; The only way is to specify the requirements and the code to execute separately and execute the code after everything has been lazy-loaded.

In order to do the same with a plugin, we should do by hand what transpilation does : collect the list of all dependencies, check them one by one (like is done in `define` here).
Amd has this advantage of "knowing the requirements on execution" - I for instance won't need the dependency grabber I PRed anymore.

But basically, there will need to be one point where the code expresses a package is needed and gets no information directly (perhaps in a callback or a promise) about the loaded package - because it is not already loaded when the call is done.

I am quite happy with the amd solution for now - of course, if you decide to implement such without amd, I would gladly come back to pure fuse-box.

PR #566, opened 2017-05-24
Export 4 all

eddow:
Hi guys,

The idea here is that, if I make an NPM package, I want the users to be able to bundle it with browserify or webpack for example.

Therefore, the package has to export its exports in `module.exports` **every time** `module.exports` is provided : not only when on the server (`!$isBrowser`) or when `__fbx__dnm__` is provided.

I added the generic export system for the npm module to be bundelable in any circumpstances (amd+cjs).


nchanged:
@eddow i have an idea. What if we introduce API style? Where we could include your solutions?

eddow:
Well,
The fact is that, before, if you executed the script, the script would have simply exposed its exports on the `window` object.
I don't know if it can be a purpose on itself - and there I don't see the difference between server and client, the global object can be modified server-side also (I don't know how/if jquery is available server-side)

Now, we could have two different syntax : one for the `globals` per se and one for the bundle exports.
Indeed, it is a bit counter-intuitive to indicate `globals: '*'` to express exports; it could be more intuitive for the bundle only to export its main-file export for instance ( +the `FuseBox` object of course)

nchanged:
thanks, I have to verify it all later on. The API is bloating up in size ;-)

eddow:
This was solved by quantum plugin

PR #567, opened 2017-05-24
Meta

eddow:
So, I am back from my `amd` solution that was a dead-end.
In order to lazy-load properly, we need a "meta" system that allows the transpiler (server-side, transpiling-time) to communicate with the executer (client-side, running time)

For this, I added the free property `meta` to `File` : this object is empty and plugins can write in it.

The client on the other hand can import meta from files with an option that will return a reference to the file and the file object (with its meta information)
```typescript
var {ref, file} = FuseBox.import(fileName, {meta: true});
```

This allows a [realy simple plugin](https://github.com/eddow/fuse-async/blob/master/src/server/DepsMetaPlugin.ts) to make available informations needed at runtime

### DIFF
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -137,7 +137,7 @@ export class BundleSource {
             || file.collection && file.collection.acceptFiles === false) {
             return;
         }
-				
+
         this.collectionSource.add(null,
             `___scope___.file("${file.info.fuseBoxPath}", function(exports, require, module, __filename, __dirname){
 ${file.headerContent ? file.headerContent.join("\n") : ""}`);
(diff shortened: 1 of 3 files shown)


nchanged:
Thanks @eddow 
You have to understand that the API is the most delicate thing in FuseBox, used in thousands of projects. Give me time to check it.

eddow:
Sure sure, I clearly understand and don't rush you; that's why I took time to check and over-check really few was modified and every change remains really generic.

Here for example, I thought it could be easy to add the dependencies in the meta-information with one line, but I thought we could let that to a plugin as many don't need this information

nchanged:
This will be closed for now.

PR #568, opened 2017-05-24
Small fixes to split bundles, source maps and hashing

bsl-zcs:
and a WIP test for bundle splitting with a separate test environment for it.

### DIFF
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -224,7 +224,7 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
             this.concat.add(null, "(FuseBox)");
         }
         if (this.context.sourceMapsProject || this.context.sourceMapsVendor) {
-            let sourceName = this.context.output.filename;
+            let sourceName = /[^\/]*$/.exec(this.context.output.filename)[0];
             this.concat.add(null, `//# sourceMappingURL=${sourceName}.js.map`);
         }
     }
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -224,7 +224,7 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
             this.concat.add(null, "(FuseBox)");
         }
         if (this.context.sourceMapsProject || this.context.sourceMapsVendor) {
-            let sourceName = /[^\/]*$/.exec(this.context.bundle.name)[0];
+            let sourceName = this.context.output.filename;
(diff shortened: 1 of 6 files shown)


src/BundleSource.ts
```diff
@@ -224,7 +224,7 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
             this.concat.add(null, "(FuseBox)");
         }
         if (this.context.sourceMapsProject || this.context.sourceMapsVendor) {
-            let sourceName = /[^\/]*$/.exec(this.context.bundle.name)[0];
+            let sourceName = this.context.output.filename;
```
nchanged: Here i see a problem.
If you do
```
bundle("foo/app")
```

Sourcemaps will be referencing a folder + filename (foo/app.js), instead of a relative path (app.js)
Please, test that

src/core/BundleSplit.ts
```diff
@@ -90,7 +90,7 @@ export class BundleSplit {
             // getting information on the paths
             configs.forEach(config => {
                 if (config) {
-                    let localFileName = config.fuse.context.output.lastGeneratedFileName;
+                    let localFileName = config.fuse.context.output.lastPrimaryOutput.filename;
```
nchanged: 👍

src/core/WorkflowContext.ts
```diff
@@ -469,7 +469,7 @@ export class WorkFlowContext {
     protected writeSourceMaps(result: any) {
         // Writing sourcemaps
         if (this.sourceMapsProject || this.sourceMapsVendor) {
-            this.output.write(`${this.bundle.name}.js.map`, result.sourceMap);
+           this.output.write(`${this.output.filename}.js.map`, result.sourceMap, true);
```
nchanged: Again, need to check with `bundle("foo/bundle")`

bsl-zcs:
> bundle("foo/app")

Well, that’s an unexpected thing to get from a field explicitly named “filename”.   :)

I’ve put the stripping regex back. It probably would be better to normalize that field in UserOutput, but it’s out of scope of the current PR.

After some patching, the manifest file for such bundle generates successfully, but relative paths to split bundles are likely wrong, at least in some corner cases.

nchanged:
Good job! Thanks a million!

nchanged:
However, I think you must have been right in the first place.
https://github.com/fuse-box/fuse-box/blob/master/src/core/UserOutput.ts#L142

So it should be working just fine, without RegExp

PR #570, opened 2017-05-25
quickfix

eddow:
```
Exception has occurred: Error
TypeError: Cannot read property 'info' of undefined
    at Function.onNode (...\fuse-box\dist\commonjs\analysis\plugins\OwnBundle.js:24:66)
```
A case of absent entry-point is not taken into account and let an `undefined.info` occur

### DIFF
--- a/src/analysis/plugins/OwnBundle.ts
+++ b/src/analysis/plugins/OwnBundle.ts
@@ -20,7 +20,7 @@ export class OwnBundle {
 
     public static onNode(file: File, node: any, parent: any) {
         const analysis = file.analysis;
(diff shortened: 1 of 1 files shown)


src/analysis/plugins/OwnBundle.ts
```diff
@@ -20,7 +20,7 @@ export class OwnBundle {
 
     public static onNode(file: File, node: any, parent: any) {
         const analysis = file.analysis;
-        if (node.type === "MemberExpression") {
+        if (file.collection.entryFile && node.type === "MemberExpression") {
```
nchanged: 👍

PR #572, opened 2017-05-25
SASS indent syntax support

iceekey:


iceekey:
Althrough CSSPlugin test fails I think it's not because of my changes. I did nothing to CSSPlugin. This changes works fine on my PC.

Maybe it fails because project on half way trough?

nchanged:
This `indentedSyntax: true` might have messed it up. Maybe we don't hardcode it and let user decide?

iceekey:
@nchanged then it should be mentioned in documentation. because without it `.sass` files can't be parsed.
`indentedSyntax: true` still allows to parse `.scss` files.

Do I need to update docs for SASSPlugin and remove `indentedSyntax: true` ? You decide.

nchanged:
Please, remove `indentedSyntax` let's leave it for the end user to decide.  Documentation - if it's easy for you, please add a few lines on additional extension support ;-) 
And thank you!

PR #579, opened 2017-05-27
adding async tag to the injected scripts

avshyz:
This one could save some hassle when working in react (namely, the notorious `Uncaught Error: _registerComponent(...): Target container is not a DOM element` error.

### DIFF
--- a/src/plugins/WebIndexPlugin.ts
+++ b/src/plugins/WebIndexPlugin.ts
@@ -49,7 +49,7 @@ $bundles
         }
 
         let jsTags = bundlePaths.map(bundle =>
-            `<script ${opts.async ? 'async' : ''} type="text/javascript" src="${bundle}"></script>`
+            `<script ${this.opts.async ? 'async' : ''} type="text/javascript" src="${bundle}"></script>`
         ).join("\n");
 
         let macro = {
--- a/src/plugins/WebIndexPlugin.ts
+++ b/src/plugins/WebIndexPlugin.ts
@@ -11,6 +11,7 @@ export interface IndexPluginOptions {
     path?: string;
     target?: string;
     template?: string;
+    async?: boolean
 }
 export class WebIndexPluginClass implements Plugin {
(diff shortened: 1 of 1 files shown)


src/plugins/WebIndexPlugin.ts
```diff
@@ -48,7 +49,7 @@ $bundles
         }
 
         let jsTags = bundlePaths.map(bundle =>
-            `<script type="text/javascript" src="${bundle}"></script>`
+            `<script ${async ? 'async' : ''} type="text/javascript" src="${bundle}"></script>`
```
nchanged: I don't see the async defined in the code, opts.async?

avshyz: it's in `IndexPluginOptions`
(I may have forgotten to add that chunk)

nchanged: OPtions I see, but async is not defined. It should be opts.async 
Sorry if I am wrong, doing the review on my phone :)

avshyz: It's not you, it's me :)
Silly mistake, duly noted and fixed!

src/plugins/WebIndexPlugin.ts
```diff
@@ -48,7 +49,7 @@ $bundles
         }
 
         let jsTags = bundlePaths.map(bundle =>
-            `<script type="text/javascript" src="${bundle}"></script>`
+            `<script ${opts.async ? 'async' : ''} type="text/javascript" src="${bundle}"></script>`
```
nchanged: Have you tested it yourself? opts is not in the scope. It's `this.opts`

avshyz: I have. yet, 33 of them have failed even before I've made my changes, and that number did not increase after.

avshyz: Sorry if that sounds half assed, It's my first open source contribution and I assumed it's just a minor tweak

PR #588, opened 2017-06-03
fix language version reference

calebboyd:


devmondo:
Thank you very much :)

PR #606, opened 2017-06-14
Don't use package browser field for non-browser targets

magnushiie:
Currently the code checks for isBrowserTarget for the overrides variant of the browser field in package.json, but does not check it for the main variant of the field.

This PR hoists the target check to the entire browser field handling.

### DIFF
--- a/src/core/PathMaster.ts
+++ b/src/core/PathMaster.ts
@@ -353,10 +353,8 @@ export class PathMaster {
                 let entryFile;
                 let entryRoot;
                 let browserOverrides;
-                if (json.browser) {
-
-                    if (this.context.isBrowserTarget() &&
-                        typeof json.browser === "object" && json.browser[json.main]) {
+                if (this.context.isBrowserTarget() && json.browser) {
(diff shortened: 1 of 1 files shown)


nchanged:
Makes sense, thanks for the PR, been thinking about the same! =)

src/core/PathMaster.ts
```diff
@@ -353,10 +353,8 @@ export class PathMaster {
                 let entryFile;
                 let entryRoot;
                 let browserOverrides;
-                if (json.browser) {
-
-                    if (this.context.isBrowserTarget() &&
-                        typeof json.browser === "object" && json.browser[json.main]) {
+                if (this.context.isBrowserTarget() && json.browser) {
```
nchanged: 👍

magnushiie:
Thanks for the quick merge!

For further information, my change was triggered by an error (`window` undefined in `useColors`) while bundling the [debug](https://www.npmjs.com/package/debug) package for server side rendering.

PR #619, opened 2017-06-18
Fix declarations for noImplicitAny

magnushiie:
Omitting the return type of fn causes compile errors in consuming projects that import fuse-box with noImplicitAny compiler option.

```log
node_modules/fuse-box/dist/typings/core/FuseBox.d.ts(63,9): error TS7020: Call signature, which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/fuse-box/dist/typings/quantum/plugin/ComputerStatementRule.d.ts(8,13): error TS7020: Call signature, which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/fuse-box/dist/typings/quantum/plugin/ComputerStatementRule.d.ts(14,13): error TS7020: Call signature, which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/fuse-box/dist/typings/quantum/plugin/QuantumCore.d.ts(30,13): error TS7020: Call signature, which lacks return-type annotation, implicitly has an 'any' return type.
```

### DIFF
--- a/src/core/FuseBox.ts
+++ b/src/core/FuseBox.ts
@@ -206,7 +206,7 @@ export class FuseBox {
 
 
     /** Starts the dev server and returns it */
-    public dev(opts?: ServerOptions, fn?: { (server: Server) }) {
+    public dev(opts?: ServerOptions, fn?: { (server: Server): void }) {
         opts = opts || {};
         opts.port = opts.port || 4444;
         this.producer.devServerOptions = opts;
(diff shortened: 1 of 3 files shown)


magnushiie:
@aretecode - the current code does not look at the return value of these functions. If it is going to look at them in the future, I think there needs to be dynamic check for this. Then the types can be changed to something like `void | string` if string was the new option to indicate that there still could be functions implemented based on the old interface.

magnushiie:
@nchanged - thanks for the quick merge, as always :)

PR #620, opened 2017-06-18
fixes #610 with uuid style browser overrides

### LINKED ISSUE #610: Browser target and deep overrides
There seems to be and issue with deep browser overrides. Issue is reproduced [here](https://github.com/erkkaha/react-example)

Problem is in this case [asap](https://github.com/kriskowal/asap/blob/master/package.json)
from asap package.json
```
  "main": "./asap.js",
  "browser": {
    "./asap": "./browser-asap.js",
    "./asap.js": "./browser-asap.js",
    "./raw": "./browser-raw.js",
    "./raw.js": "./browser-raw.js",
    "./test/domain.js": "./test/browser-domain.js"
  }
```

"./asap.js" --> "./browser-asap.js" works fine, but "./browser-asap.js" requires  "./raw" and that is not resolved to "./browser-raw.js".

Same issue can be reproduced with uuid as well, where index.js requires v4.js, which requires ./lib/rng.js, which should resolve to "./lib/rng-browser.js"

Tested with fuse-box 2.1.0 and node 8.0

erkkaha:
These changes enable browser target even when main entry point is not included in the browser overrides,
fixes also issue, when require statement is without file extension like in uuid `var rng = require('./lib/rng');` and package.json browser is with file extensions `  "browser": {
    "./lib/rng.js": "./lib/rng-browser.js"
  },`

### DIFF
--- a/src/analysis/plugins/ImportDeclaration.ts
+++ b/src/analysis/plugins/ImportDeclaration.ts
@@ -70,8 +70,10 @@ export class ImportDeclaration {
         if (file.collection && file.collection.info && file.collection.info.browserOverrides) {
             const overrides = file.collection.info.browserOverrides;
             if (overrides) {
-                if (overrides[requireStatement]) {
-                    requireStatement = overrides[requireStatement];
+                //require statement without file ext and override with ext
+                let requireStatementWithExt = requireStatement + '.js';
+                if (overrides[requireStatement] || overrides[requireStatementWithExt]) {
+                    requireStatement = overrides[requireStatement] || overrides[requireStatementWithExt];
                     file.analysis.requiresRegeneration = true;
                 }
             }
--- a/src/core/PathMaster.ts
+++ b/src/core/PathMaster.ts
@@ -354,9 +354,11 @@ export class PathMaster {
                 let entryRoot;
                 let browserOverrides;
                 if (this.context.isBrowserTarget() && json.browser) {
-                    if (typeof json.browser === "object" && json.browser[json.main]) {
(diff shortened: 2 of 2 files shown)


src/analysis/plugins/ImportDeclaration.ts
```diff
@@ -70,8 +70,10 @@ export class ImportDeclaration {
         if (file.collection && file.collection.info && file.collection.info.browserOverrides) {
             const overrides = file.collection.info.browserOverrides;
             if (overrides) {
-                if (overrides[requireStatement]) {
-                    requireStatement = overrides[requireStatement];
+                //require statement without file ext and override with ext
+                let requireStatementWithExt = requireStatement + '.js';
```
aretecode: this extension should either be configurable, or allow a function callback, or try a require.resolve

nchanged: Disagreed, it's a package json thingy with "browser" field

nchanged: However, a callback might be helpful

src/analysis/plugins/ImportDeclaration.ts
```diff
@@ -70,8 +70,10 @@ export class ImportDeclaration {
...
+                if (overrides[requireStatement] || overrides[requireStatementWithExt]) {
+                    requireStatement = overrides[requireStatement] || overrides[requireStatementWithExt];
```
nchanged: 👍

PR #634, opened 2017-06-25
feat: allow traversal plugins to be passed through makeAnalysis

calebboyd:
for when a plugin needs to manipulate or traverse the AST

### DIFF
--- a/src/analysis/FileAnalysis.ts
+++ b/src/analysis/FileAnalysis.ts
@@ -11,7 +11,12 @@ import { DynamicImportStatement } from "./plugins/DynamicImportStatement";
 require("acorn-es7")(acorn);
 require("acorn-jsx/inject")(acorn);
 
-const plugins: any = [AutoImport, OwnVariable, OwnBundle, ImportDeclaration, DynamicImportStatement];
+export interface TraversalPlugin {
+    onNode(file: File, node: any, parent: any): void
+    onEnd(file: File): void
+}
+
+const plugins: TraversalPlugin[] = [AutoImport, OwnVariable, OwnBundle, ImportDeclaration, DynamicImportStatement];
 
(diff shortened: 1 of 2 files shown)


src/analysis/FileAnalysis.ts
```diff
@@ -123,18 +123,23 @@ export class FileAnalysis {
         return node.type === "Literal" || node.type === "StringLiteral";
     }
 
-    public analyze() {
+    public analyze(traversalOptions?: { plugins: any[] }) {
```
nchanged: Should be typed, please add it

src/core/File.ts
```diff
@@ -298,11 +298,11 @@ export class File {
         this.isLoaded = true;
     }
 
-    public makeAnalysis(parserOptions?: any) {
+    public makeAnalysis(parserOptions?: any, traversalOptions?: { plugins: any[] }) {
```
nchanged: Same here, typings (cuz we have them)

calebboyd: Are you suggesting I create a "`TraversalPlugin`" interface? (i don't see one currently)
They're just typed as any right now..
https://github.com/calebboyd/fuse-box/blob/master/src/analysis/FileAnalysis.ts#L14

```typescript
interface TraversalPlugin {
    onNode(file: File, node: any, parent: any): void
    onEnd(file: File): void
}
```

calebboyd: Added a commit w/ ^. Just let me know if it LGTY

nchanged: Awesome! thanks a lot! We didn't have it, now we do ;-)

calebboyd:
Thanks!

nchanged:
The next beta is tomorrow, your changes will come through

PR #642, opened 2017-06-28
Add charset/description/keywords/author metas to WebIndexPlugin config

otothea:
Add optional config props for common html metas.

```
WebIndexPlugin({
  charset: 'utf-8',
  title: 'Document Title',
  description: 'A description of the web app',
  keywords: 'keywords, that, represent, your, app',
  author: 'Guy <guy@author.com>'
})
```

I also slightly modified the $title template variable to wrap the `<title>` tag for you, instead of having to include it in the template yourself.

I also fixed a bug in the default html template where the `$bundles` var is outside of the `<body>` tag.

### DIFF
--- a/src/plugins/WebIndexPlugin.ts
+++ b/src/plugins/WebIndexPlugin.ts
@@ -6,6 +6,10 @@ import { ensureAbsolutePath, joinFuseBoxPath } from "../Utils";
 
 export interface IndexPluginOptions {
     title?: string;
+    charset?: string;
+    description?: string;
+    keywords?: string;
+    author?: string;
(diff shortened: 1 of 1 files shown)


otothea:
Build failure seems to be a system error

`Error: watch /home/travis/build/fuse-box/fuse-box/src/tests/stubs/sparky/foo ENOSPC`

nchanged:
These are breaking changes, I would suggest keeping title as it was, and add some extra variable for convenience, like &htmlTitle

otothea:
@nchanged Reverted changes to the $title var

PR #664, opened 2017-07-12
Add babel support to VuePlugin

Zhomart:
Added tests and docs updated.


nchanged:
That's some kick ass PR. 👍  I will check it out soon!

PR #667, opened 2017-07-13
Update bundle plugin docs to match implementation. Fix #666

### LINKED ISSUE #666: Documentation incorrectly states that plugins will be overridden by .plugin() calls
[The bundle documentation for overrides](http://fuse-box.org/page/bundle#option-override) says that the following code will have a final value for plugins that is: `plugins: [CSSPlugin()]`, when it will actually be `plugins: [HTMLPlugin(), CSSPlugin()]`.
```
const fuse = FuseBox.init({
    homeDir: "src",
    output: "dist/$name.js",
    plugins : [HTMLPlugin()]
});

fuse.bundle("bundle1")
      .plugin(CSSPlugin())
      .instructions(`~ index.ts`);
```

nspaeth:
Since plugins actually append I updated the example to show that, and also
included an example where the option gets completely overwritten.

(It looks like I accidentally also deleted a bunch of trailing whitespace. If that is a problem I can fix it).

### DIFF
--- a/docs/bundle.md
+++ b/docs/bundle.md
@@ -24,7 +24,7 @@ graph TB
     main --> run((Execute bundles))
 ```
 
-You can create infinite amount of bundles as well as infinite amount of `Producers`. 
+You can create infinite amount of bundles as well as infinite amount of `Producers`.
 
 ## Producer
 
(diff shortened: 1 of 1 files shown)


nchanged:
Awesome! Thanks

PR #668, opened 2017-07-14
fix css url reconstruction

devmobileaim:
CSSResourcePlugin was putting double parenthesis in url reconstruction


nchanged:
thanks!

PR #669, opened 2017-07-14
Fix #665 - make all plugins apply to matching files instead of just the first matching plugin

### LINKED ISSUE #665: only the first plugin to match a file will run on it
If, for example, you have `plugins: [ReplacePlugin(...), ReplacePlugin(...)]`, only the first plugin will run on any file, not executing the second transform. This is problematic if, for example, you have the TypeScriptHelpers() plugin before your ReplacePlugin(), or if you add a second ReplacePlugin() later on in one of your tasks.

For example, I have tasks: 
```
Sparky.task('config', [], () => {
  fuse = new FuseBox({
    plugins: [ReplacePlugin({ __PRODUCTION__: isProduction })]
  })
})

Sparky.task('ios', ['config'], () => {
 ios = fuse
  .bundle('ios')
  .plugin(ReplacePlugin({__IOS__: true}))....
})
```
Where only `__PRODUCTION__` gets replaced, and not `__IOS__`.

In `tryPlugins()` in File.js you can see that it matches the first plugin then exits the list, instead of executing all matching plugins or queuing them.

nspaeth:


nchanged:
That's a breaking change, I don't think we can do that.

nspaeth:
The existing behavior is relatively unexpected (and limiting), so it would be nice to eventually change it.

Are there any other planned breaking changes that this could be grouped with for a later version?

nchanged:
Yes with FuseBox 3 we will redesign the entire API these change could come, however I need to investigate it. This PR might break some people's builds, let's keep it open for now

nspaeth:
See new pull request #677

PR #671, opened 2017-07-15
fix: fusebox bundle in repl and _third_party_main

calebboyd:
This allows a fusebox bundle to be loaded into the environment where `global.require.main` isn't set

This occurs in at least two scenarios: Node REPL and when [_third_party_main.js](https://github.com/nodejs/node/blob/v8.1.4/lib/internal/bootstrap_node.js#L87) is used.

Maybe it would be best to leave it undefined. I'm not sure what the role of main is throughout the rest of the API


modules/fuse-box-loader-api/fusebox.js
```diff
@@ -276,10 +276,12 @@ function $import(name, o) {
             v: ref.versions,
         });
     };
-    locals.require.main = {
-        filename: $isBrowser ? "./" : g["require"].main.filename,
-        paths: $isBrowser ? [] : g["require"].main.paths,
-    };
+    if ($isBrowser || !g["require"].main) {
```
nchanged: `g["require"]` is only working on the server, don't you think we can revise that?

calebboyd: I'm not sure I follow your question. 
`g["require"]` only gets evaluated if `$isBrowser` is not truthy. Is that check not sufficient?

nchanged: RIght, sorry i misread it!

PR #677, opened 2017-07-19
Fix #665 - make all plugins apply to matching files instead of just the first matching plugin 

### LINKED ISSUE #665: only the first plugin to match a file will run on it

nspaeth:


src/core/File.ts
```diff
@@ -221,6 +221,7 @@ export class File {
      * @memberOf File
      */
     public tryPlugins(_ast?: any) {
+        if (this.context.runAllMatchedPlugins) { this.tryAllPlugins(_ast) }
```
nchanged: I think in this case you will have them both run (allPlugins + the current scenario). I don't think it's correct

nspaeth: You're right, there was meant to be a return there. I've pushed a change to fix that.

PR #703, opened 2017-08-02
Suggest node_modules as path to pass to Less

troynt:
In Webpack one would use
```css
@import "~foo/dist/example.less";
```
See https://github.com/webpack-contrib/less-loader#imports

We should encourage users to provide node_modules path to plugin to avoid confusion around import statements.

Better, would be to adjust LESSPlugin to resolve paths prefixed by ~.

### DIFF
--- a/docs/plugins/css/LessPlugin.md
+++ b/docs/plugins/css/LessPlugin.md
@@ -27,7 +27,12 @@ Inject into a chain.
 
 ```js
 fuse.plugin(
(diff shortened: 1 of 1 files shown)


nchanged:
Are you sure that source maps don't work?

troynt:
@nchanged I just copied the note that was already there, I haven't tested source maps. Source maps are broken in my project, but I'm not sure if my configuration is to blame.

Issue #699, opened 2017-08-01
Angular AOT and alternate typescript compilers


calebboyd:
Please correct me if I missed it, but is there a mechanism to use an alternate typescript compiler? 

Such as byots or angular's ngc? Right now it seems like the default `typescript` compiler is hardcoded.

Wondering if it would make sense to put it in a plugin. But would need a way to disclude it if you wanted to provide your own...

devmondo:
@calebboyd  Fuse uses TS by default, but nothing prevents you from creating a plugin to plug in your favorite compiler, the same we did for babel, and even coffeescript.

calebboyd:
As far as I can tell I cannot prevent fusebox from trying to transpile typescript. (and do it myself instead). For example, if I want to use the AOT angular compiler.

If I try to make a `.ts` plugin and use a different one I inevitably end up with:

```
{ Error: Cannot find module 'typescript'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at File.handleTypescript (/my/project/node_modules/fuse-box/dist/commonjs/core/File.js:283:20)
```

calebboyd:
If i'm reading it right I can apply a plugin to a typescript file, but only if the typescript file is already cached??
https://github.com/fuse-box/fuse-box/blob/master/src/core/File.ts#L466

I could also use `onTypescriptTransform` but that still does double work if I wanted to transpile it myself.

nchanged:
@calebboyd yes we could have an option that would override and give a full control. But there are places where typescript is mandatory as, for example, it transpiles node_modules if es6 is detected

calebboyd:
https://gitter.im/fusebox-bundler/Lobby?at=5981e55b4bcd78af56098b63
```
@calebboyd 09:44
Why would node modules be es6?
to support treeshaking? -- if so what will typescript do anyway if its configured to output es2015 modules

@nchanged 09:45
that’s for node_modules
if there is an es6 module ts will transpiled it
@troynt feel free to improve the plugin, but don’t make breaking changes please https://github.com/fuse-box/fuse-box/blob/master/src/plugins/WebIndexPlugin.ts

@nchanged 09:46
to support running on fusebox -)

@calebboyd 09:48
I'm not following... Why would someone publish an unconsumable module on npm

@nchanged 09:48
angular

calebboyd @calebboyd 09:50
[...]

PR #704, opened 2017-08-02
feat: add typescript bypass

calebboyd:
This is a follow up to #699. 

I still have to see If I can get angular AOT setup with this. But thats my general idea.


src/core/File.ts
```diff
@@ -117,6 +117,11 @@ export class File {
 
     public groupHandler: Plugin;
 
+    /**
+     * Flag to indicate to fusebox whether or not contents has been set to consumable javascript
+     */
+    public transpiled: Boolean = false;
```
nchanged: I would rename this one to `isTypescriptTranspiled` or maybe some better name, can't just figure out what would that be.  cuz `transpiled` is quite a generic name, we do however a first class typescript support. 

Could you create some documentation in the plugin section to help people understand and make use of this option?

calebboyd: Yeah, I can do that. Thanks. I mostly made this to illustrate my goal. I'll refine it more once I validate it is a valid approach

PR #706, opened 2017-08-03
import as exported

eddow:
Little change, big effects.

When a vue single-file component is defined, it exports the component as default :
```
@Component
export default class myComponent extends Vue {
```

So, it would seem logical to import it as default too :
```
import myComponent from '../components/myComponent.vue'
```

Instead, for now, the default has become the "whole" export, so we have to use this instead :
```
import * as myComponent from '../components/myComponent.vue'
```

Allowing it to remain `default`, beside notation consistency, would also allow component single-files to export other values, like global configurations, etc.

### DIFF
--- a/src/plugins/VuePlugin.ts
+++ b/src/plugins/VuePlugin.ts
@@ -144,7 +144,7 @@ var _v = function(exports){${jsContent}
 _p.render = ` + toFunction(compiled.render) + `
 _p.staticRenderFns = [ ` + compiled.staticRenderFns.map(toFunction).join(',') + ` ];
 var _e = {}; _v(_e); Object.assign(_e.default.options||_e.default, _p)
-module.exports = _e.default
+module.exports = _e
     `;
 }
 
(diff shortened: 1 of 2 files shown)


nchanged:
tests have failed;-)

eddow:
Hi again,

So, I changed the tests because, as I said - even if it is more logical it is a different notation for usage.
What do you think about this change for now?

nchanged:
@eddow I just don't feel Ok with exporting `default` by default.

eddow:
Well, this is the generic way of doing in typescript modules.
```
export default var myVar = 42;
export var auxVar = 29;
```
calls for :
```
import myVar, {auxVar} from 'myLib'
```

if you use `import * as myValues from 'myLib'`, `myValues` will be :
```
{
 default: 42,
 auxVar: 29
}
```

an `export default class myComponent` calls for `import myComponent from...`

nchanged:
So it follows the standard, if you export a default value you will get it as `default` if you don't you get it as is, right?

eddow:
Well, the `VueImport` plugin didn't follow the standard, that is why I made the PR.

As it was done,
```
export default class myComponent
```
was calling for
```
import * as myComponent from ...
```

My PR is about putting it back to
```
import myComponent from...
```

PR #713, opened 2017-08-08
allow uglify options via quantum

dgreif:
By checking specifically for `true` here, it prevents an uglify options object from being passed through Quantum.  

https://github.com/fuse-box/fuse-box/blob/master/src/quantum/plugin/BundleWriter.ts#L16 is already set up to take uglify options from this method.

### DIFF
--- a/src/quantum/plugin/QuantumOptions.ts
+++ b/src/quantum/plugin/QuantumOptions.ts
@@ -196,7 +196,7 @@ export class QuantumOptions {
     }
 
     public shouldUglify() {
-        return this.uglify === true;
+        return this.uglify;
     }
 
     public shouldBakeApiIntoBundle() {


nchanged:
Quantum has an independent uglify-js configuration, why would you want to have it like this?

calebboyd:
If you can't pass options like ascii_only, quantum will mangle your output in some cases.

nchanged:
right,I will check how we can pass the options to uglify-js in Quantum

calebboyd:
I think this change allows that.

nchanged:
That's right!

dgreif:
Thanks @nchanged! When do you plan to release the next beta?

nchanged:
tomorrow

cfatorNSS:
I tried out the Release 2.2.2-beta.11 but do not see how options can be passed to uglify.

dgreif:
We are doing something like this:

```
QuantumPlugin({
  uglify: {
    compress: { warnings: true },
    output: { comments: false, 'ascii_only': true },
    sourceMap: true
  },
  treeshake: true
})
```

You simply pass an object to `uglify` instead of `true`

PR #718, opened 2017-08-09
feat: centralized error logging for bundles

dgreif:
I decided to store the errors on the `Bundle`s instead of `BundleProducer`.  This allows us to clear any existing errors each time a bundle executes, without having to worry about other bundles that are/are not executing at the same time.  All current errors across bundles can be accessed through `producer.getErrors()`

I updated the `SassPlugin` to use the new error logging, and update the output to be more useful.

Once this is in, we can discuss the best path for getting the errors to the browser.

image

### DIFF
--- a/src/Log.ts
+++ b/src/Log.ts
@@ -359,6 +359,9 @@ export class Log {
         log.red().bold(msg).echo();
         return this;
     }
+    public echoError(str: string) {
+        log.red(`  → ERROR ${str}`).echo()
+    }
     public echoRed(msg) {
         log.red(msg).echo();
         return this;
--- a/src/core/Bundle.ts
+++ b/src/core/Bundle.ts
@@ -26,6 +26,7 @@ export class Bundle {
     public lastChangedFile: string;
     public webIndexed = true;
     public splitFiles: Map<string, File>;
+    private errors: string[] = [];
 
     public bundleSplit: BundleSplit;
     public quantumItem: QuantumItem;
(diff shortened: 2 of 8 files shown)


dgreif:
I had some free time tonight, so I went ahead and implemented the browser logging as well.  It will output the errors in both the console and the DOM.  I threw some styles on, but I'm happy to change it to look however you want.  It will create a fixed div and display the errors over top of the application, removing the div when there are no longer any errors.

image

nchanged:
@dgreif such great job man!!! I love it.
I didn't have much to play with it, but I have merged it into a new branch `error-logging` where we could continue working on it.

The way I see the reporting in the browser should be less intrusive. Let's say, one could fold it into a small strip `50x200px` on the right/left-hand side of the browser window. A small button will do. Once an error occurred it would display a status `Sass Error` and turn red (obviously). We also would need to save user preference (folder/unfolded) in local storage, so after a refresh, an error wouldn't come straight to the user's face.


Would do you think?

P.S Thanks a lot for your contribution, let's continue polishing it!

Shepless:
What's the status of this? Looks incredibly useful!

nchanged:
It’s merged 1 month ago xD

TombolaShepless:
Linked PR is still open? Was it a diff PR?

nchanged:
@TombolaShepless you are in the PR ;-)

TombolaShepless:
OMG - far too early in the morning, clear my brain hasn't woken up with the rest of my body yet.... :man_facepalming:

PR #726, opened 2017-08-14
Quantum "npm" target

eddow:
When working with modules that are independently used server-side and client-side, there is a need of creating a module file that uses the standard require/module.exports, regardless of its usage on the server or in the browser.

I corrected here 2 little points about the "universal" target that, I think, was done for this figure case :
- The values exported by the fused module are given in module.exports (like in the case of "server" target)
- The unresolved imports are just imported with the `require` keyword.

This way, I could make a module that just exports functions, whether it is bundled for server or browser use.

### DIFF
--- a/src/quantum/plugin/FlatFileGenerator.ts
+++ b/src/quantum/plugin/FlatFileGenerator.ts
@@ -76,7 +76,7 @@ export class FlatFileGenerator {
                     this.contents.push(`module.exports = ${req}`);
                 }
 
-                if (this.core.opts.isTargetUniveral()) {
+                if (this.core.opts.isTargetNpm()) {
                     this.contents.push(`module.exports = ${req}`);
                 }
 
--- a/src/quantum/plugin/FlatFileGenerator.ts
+++ b/src/quantum/plugin/FlatFileGenerator.ts
@@ -77,7 +77,7 @@ export class FlatFileGenerator {
                 }
 
                 if (this.core.opts.isTargetUniveral()) {
-                    //this.contents.push(`module.exports = ${req}`);
+                    this.contents.push(`module.exports = ${req}`);
                 }
 
                 if (this.core.opts.isTargetBrowser()) {
--- a/src/quantum/plugin/FlatFileGenerator.ts
+++ b/src/quantum/plugin/FlatFileGenerator.ts
@@ -72,11 +72,7 @@ export class FlatFileGenerator {
             const req = `$fsx.r(${JSON.stringify(this.entryId)})`;
 
             if (this.globalsName) {
-                if (this.core.opts.isTargetServer()) {
-                    this.contents.push(`module.exports = ${req}`);
-                }
-
-                if (this.core.opts.isTargetNpm()) {
+                if (this.core.opts.isTargetNpm() || this.core.opts.isTargetServer()) {
                     this.contents.push(`module.exports = ${req}`);
                 }
 
--- a/src/quantum/plugin/QuantumOptions.ts
+++ b/src/quantum/plugin/QuantumOptions.ts
@@ -222,7 +222,10 @@ export class QuantumOptions {
         return this.optsTarget === "electron";
     }
     public isTargetUniveral() {
-        return this.optsTarget === "universal";
+        return this.optsTarget === "universal" || this.optsTarget === "npm";
+    }
+    public isTargetNpm() {
+        return this.optsTarget === "npm";
     }
 
     public isTargetServer() {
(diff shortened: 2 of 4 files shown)


eddow:
Well, I don't see any use-case where `preserveRequireStatements: false` would be useful, but I can understand though.

Beside, I never really modified the options of quantum - is there any caveats I should take care for if I try?

nchanged:
@eddow Quantum creates an independent environment agnostic build. `require` is a global function which is not available in the browser.

eddow:
The purpose is to build a file that can be bundled.
1- Quantum takes care of if it is used with node or not, so it is already not environment agnostic
2- There are 3 cases of usage : 
 a- on the server with node
 b- in a browser by exposing the library in the globals
 c- environment-agnostic letting a bundler/systemjs/requirejs/whatever manage the requirements
Basically, A and C are the same : it's the commonJS standard. The commonJS standard is used in the libraries for browser-side too, and has been used in order to make libraries environment agnostic.
For now, fuse-box has troubles bundling commonJS libraries for the browser.

nchanged:
@eddow 
`For now, fuse-box has troubles bundling commonJS libraries for the browser.`
What do you mean? What troubles? Of course it bundles it!! 
I am not going to merge this PR until you read and understand what I am saying. There will be no `require` statements in the bundle output, as the browser doesn't have `require` globally, and I am not going to pollute window object with it.

eddow:
So, as we spoke, I added the new `target: "npm"`
I did put back the `universal` one to exactly how it behaved before and it had been tested on an npm project

src/quantum/plugin/QuantumOptions.ts
```diff
@@ -222,7 +222,10 @@ export class QuantumOptions {
         return this.optsTarget === "electron";
     }
     public isTargetUniveral() {
-        return this.optsTarget === "universal";
+        return this.optsTarget === "universal" || this.optsTarget === "npm";
```
nchanged: I don't think it's the same, it will resolve in conflict

src/quantum/plugin/FlatFileGenerator.ts
```diff
@@ -76,8 +76,8 @@ export class FlatFileGenerator {
                     this.contents.push(`module.exports = ${req}`);
                 }
 
-                if (this.core.opts.isTargetUniveral()) {
-                    //this.contents.push(`module.exports = ${req}`);
+                if (this.core.opts.isTargetNpm()) {
```
nchanged: Should be combined with target server, as it's the same

src/quantum/plugin/modifications/StatementModifaction.ts
```diff
@@ -41,8 +41,9 @@ export class StatementModification {
                     statement.setFunctionName('$fsx.r');
                     statement.setValue(resolvedFile.getID());
                 } else {
-
-                    if (core.opts.isTargetServer() || core.opts.isTargetUniveral()) {
+                    if(core.opts.isTargetNpm()) {
+                        statement.setFunctionName('require');
```
nchanged: Please test it, because there is no need in settings to `require`, as we are transforming from `require`. Long story short, no changes to the statement function

src/quantum/plugin/QuantumOptions.ts
eddow: Well, in general, "npm" target has the same behaviour as "universal" target. So, unless in a (only one indeed) specific case, everytime the target is tested, an "npm" target should lead to the same behaviour of an "universal" target.
This changes affect the way things are managed when `target: "npm"`, not when `target: "universal"`

src/quantum/plugin/modifications/StatementModifaction.ts
eddow: An `npm` module uses the `require` statement to access external (vendor)  packages that are not in the bundle. If fuse-box removes all the `require`, it won't be able to bundle a package that uses other npm packages without bundeling them.

The internal `require`-s are transformed into `$fsx.s` but the ones that refer to external packages have to remain as `require`

src/quantum/plugin/QuantumOptions.ts
nchanged: That's not correct, universal - means it will be working on browser too, where there is no `require`

eddow: When `target: "universal"`, `isTargetNpm` evaluates to `false`.

The statement `require` will be used when and only when `isTargetNpm` evaluates to `true`.

Therefore, when `target: "universal"`, the statement used will be `$fsx.s`, like expected

nchanged: Ok, lemme test it manually (it's quite an important addition), I will get back to you, later on, today, thanks!

Issue #734, opened 2017-08-17
VuePlugin : usage of the <style> of single-file components


eddow:
The `vue-template-compiler` returns all the parts of a single-file `.vue` component : script, template and styles.

For now, the styles are not taken into account - to complete the plugin would mean to manage the given styles in the compiled file with later steps to have a complete management of .vue styling:
- style scopes (note: I opened a [thread on vue forum](https://forum.vuejs.org/t/single-file-components-style-and-scope-types/15972) to clarify some details of that part)
- style transpilation (to use CSS-preprocessors)

PR #741, opened 2017-08-21
Vue styling

eddow:
https://github.com/fuse-box/fuse-box/issues/734
So again - as I need it I begun it.
Here is the first shot for Vue styling.
The CSS are loaded like other CSS files with `fuse-box-css` when the Vue file is loaded.

For now, only the un-scoped CSS is implemented (scoped and transpiled are not) so that we have time to chat about how & details.

One detail to know about transpilation is that I found only one generic transpiler for CSS : [accord](https://github.com/jescalan/accord)
It does quite the same job as `consolidate` except that consolidate doesn't deal with CSS pre-processors.

If you know of another, please say. If I use `accord`, perhaps it will be nice to replace what has been done with `consolidate`  afterward (I let you see)

### DIFF
--- a/src/plugins/VuePlugin.ts
+++ b/src/plugins/VuePlugin.ts
@@ -13,6 +13,7 @@ let vueTranspiler;
 let typescriptTranspiler;
 let babelCore;
 let babelConfig;
+
 export class VuePluginClass implements Plugin {
     public test: RegExp = /\.vue$/;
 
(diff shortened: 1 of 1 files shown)


src/plugins/VuePlugin.ts
```diff
@@ -130,17 +145,28 @@ function compileBabel(options, context, html, script) : string {
...
+function reduceVueToScript(file, jsContent, html, styles) : string {
     const compiled = vueCompiler.compile(html);
+    var cssInclude = '';
+    if(styles.global)
+        cssInclude += styles.global;
+    if(styles.scoped)
+        console.error('Functionality not yet supported: scoped style')
+    if(cssInclude) cssInclude =
+        'require("fuse-box-css")('+
+        JSON.stringify(file.info.fuseBoxPath)+
```
nchanged: It will include the current path, isn't it wrong? e.t `components/Hello.vue`

nchanged: Oh sorry, i see now

PR #746, opened 2017-08-22
Target npm

eddow:
I had a bug report on a line appearing in the npm bundle :
```
global.require = require
```
In an npm bundle, this definitely makes no sense. The modification basically removes this line allowing as a side effect `fuse-box-responsive-api` to access the `npm` target

### DIFF
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -8,6 +8,7 @@
     /* @end */
 
     /* @if universal */
+    var isBrowser = typeof window !== "undefined";
     /* @if !npm */
     var isBrowser = typeof window !== "undefined";
     if (!isBrowser) {
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -9,11 +9,6 @@
 
     /* @if universal */
     var isBrowser = typeof window !== "undefined";
-    /* @if !npm */
-    if (!isBrowser) {
-        global.require = require;
-    }
-    /* @end */
     /* @if !isContained */
     var storage = isBrowser ? window : global;
     if (storage.$fsx) {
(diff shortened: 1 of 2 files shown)


nchanged:
You sure there will be no issues with local `require` function in some fusebox local modules?

eddow:
With this : `/* @if !npm */`, yes : because `/* @if npm */`, `require` is provided by the commonjs engine using the bundle

modules/fuse-box-responsive-api/index.js
```diff
@@ -9,10 +9,11 @@
 
     /* @if universal */
     var isBrowser = typeof window !== "undefined";
+    /* @if !npm */
     if (!isBrowser) {
```
nchanged: well here, npm means `no browser` by default, or universal has to be there? Because `npm` should be like extended target `server`, but not `universal`

eddow: An npm library can be used server-side (with node) or client-side (bundled with fuse-box, webpack, etc.)
The only common part is that the engine using the npm library follows the `commonjs` standards and provide a `require` function to the library.

`commonjs` allows the "plug-n-play" saying that each libraries `require` other libraries and `module.exports` or `exports` values.

So, a npm library doesn't have to provide a `require` as the `require` will be provided by the engine (node/bundler/...), and this is valid agnostically of server/client side

nchanged: Exactly my point, target : `npm` and `client` and `universal` are incompatible. Publishing will npm will preserve require statements, which will handled accordingly by a bundling tool.

eddow: yes, so if target is `npm`, there should not be any `global.require = require` - that is what I removed here, saying that is line is used only when the target is not npm.

eddow:
Hm, I don't understand; you can see in my changes no reason for `gulp dist` to fail and it doesn't fail locally - any idea of why it could have stopped building ?

nchanged:
I will deal with it don't worry

nchanged:
As it has been fixed, I am closing this PR now

PR #759, opened 2017-08-29
feat(error-logging): minimize/hide in browser

dgreif:
## Updates
* Send bundle errors to new browsers as soon as they open.  Previously they would only get the errors in the browser after a consecutive bundle execution.
* Allow user to minimize the error output in the browser
  * This is persisted in local storage, so output stays minimized after refresh
  * Minimized:
image
  * Expanded:
image
* Allow user to hide the error output
  * This is not persisted in local storage.  If a user doesn't want errors in the browser, they can set `showErrorsInBrowser: false`

Big thanks to @adye for putting together the styles 🎉

### DIFF
--- a/src/core/Bundle.ts
+++ b/src/core/Bundle.ts
@@ -91,18 +91,29 @@ export class Bundle {
             });
 
             if (this.context.showErrorsInBrowser) {
+                const type = "update-bundle-errors",
+                    getData = () => ({
+                        bundleName: this.name,
+                        messages: this.errors
+                    })
+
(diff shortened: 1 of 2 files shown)


nchanged:
Wow! impressive! I will merge it now and check for myself. We need to hook as many error as possible

dgreif:
Agreed.  I was hoping you could re-work the plugins to register errors.  The Sass plugin was pretty simple, but you know all the other plugins way better than I do.

PR #782, opened 2017-09-06
feat: expose `package.json` in file info

calebboyd:
This sets the raw `package.json` data onto a `packageJson` property in `IPackageInformation`

Also gets the sass dep extraction tests working on windows.

The test for the `packageJson` property is a bit hacky, mostly because there is no way (afaik) to configure a PathMaster instance outside of the global Config object

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "url": "git+https://github.com/fuse-box/fuse-box.git"
   },
   "author": "",
-  "license": "ISC",
+  "license": "MIT",
   "bugs": {
     "url": "https://github.com/fuse-box/fuse-box/issues"
   },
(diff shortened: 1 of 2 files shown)


nchanged:
Do you really think we need to carry `package.json` data everywhere?

calebboyd:
I figured its useful contextual information for a file. Its just a json object. If you don't to merge it thats fine. I suppose the plugin could just pull it from the nearest node_modules folder of an "entryFile"...

I've removed it -- still have the other things in though if you want to merge them.

nchanged:
i was just thinking, since it's in memory already it can be in the object, I think it's fine.
Btw it won't be available in every file, just the entry point.

calebboyd:
Ahh, I was thinking I needed to identify each module to find the native modules... but I think it can just inspect the dependencies as they become available.

PR #786, opened 2017-09-07
fix: acorn 'script' for parsing non-strict input

### LINKED ISSUE #785: Acorn "script" vs "module" -- allow configuration?
I'm wondering if i missed it somewhere. -- Allow bundling of non-strict code.

Any way to pass a different configuration to acorn?

calebboyd:


nchanged:
Oh I tried type script, it was failing badly

calebboyd:
Hmm.. What do you think a good solution would be? -- It looks like is occurring after parsing?... Is that required?

Issue #778, opened 2017-09-04
TS/SassPlugin and SCSS imports have different paths in SourceMaps file


rs3d:
### Steps to reproduce:

1. in `app/index.ts` => `import('index.scss');`
2. in `app/index.scss` => `@import 'sub';`
3. use following setup: 

```
const {
  FuseBox,
  SassPlugin,
  CSSPlugin,
} = require('fuse-box');

const path = require('path');

const basePath = 'app';
const distPath = 'dist';

const fuse = FuseBox.init({
  homeDir: basePath,
  output: `${distPath}/$name.js`,
  tsConfig: 'tsconfig.json',
  log: true,
  debug: true,
  cache: false,
  sourceMaps: {
    inline: false,
    sourceRoot: `/${basePath}`
  }
});

fuse.bundle(basePath)
  .plugin([
    SassPlugin({
      sourceMap: true,
      outputStyle: 'compressed'
    }),
    CSSPlugin({
      inject: file => `/${file}`,
      outFile: file => `${distPath}/${file}`
    })
  ])
  .target('browser')
  .instructions(`> index.ts`);

fuse.run();
```
### Demo:
https://github.com/rs3d/fuse-box-sass-issue

### Results in `dist/index.scss.map`
```
{"version":3,"file":"index.scss","sources":["/app/index.scss","/app/app/_sub.scss"],"names":[],"mappings":"ACAA,AAAA,IAAI,AAAC,CACD,UAAU,CAAE,GAAG,CAClB,ADAD,AAAA,IAAI,AAAC,CACD,KAAK,CAAE,IAAI,CACd"}
```
==> wrong path here `/app/app/_sub.scss`

rs3d:
Forgot to mention that i used FuseBox with this PR https://github.com/fuse-box/fuse-box/pull/773/files
Without this change `_sub.scss` is not included as source.

zigmo:
@rs3d - if you move `_sub` one level down e.g. `styles/_sub.scss` helps right? This is what I was thinking of as I created the PR. I tried to avoid any breaking changes of this [line](https://github.com/fuse-box/fuse-box/blob/813bc793c7d9de764a5f9aac64889233426f2c0a/src/core/File.ts#L594)  - ofc I don't know if its a breaking change.

[...]

Issue #543, opened 2017-05-17
Using UglifyJS with sourceMap is not handled by the browser


RPDeshaies:
Here's the fusebox Config : 

```
const fuse = fsbx.FuseBox.init({
  homeDir: './client/src',
  output: './client/build_dev/static/$name.js',
  tsConfig: './tsconfig.json',
  sourceMaps: true,
  shim: shims,
  plugins: [
    fsbx.EnvPlugin({
      NODE_ENV: process.env.NODE_ENV
    }), [fsbx.SassPlugin({
      outputStyle: dev ? 'nested' : 'compressed'
    }), fsbx.PostCSS([autoprefixer({ remove: false, browsers: ['last 6 version'] })]), fsbx.CSSPlugin()],
    fsbx.BabelPlugin({
      comments: false,
      presets: ['es2015']
    }),
    fsbx.TypeScriptHelpers(),
    fsbx.JSONPlugin(),
    fsbx.CSSPlugin(),
    fsbx.RawPlugin(['*.svg']),
    fsbx.HTMLPlugin({
      useDefault: false
    }), fsbx.UglifyJSPlugin({
      compress: {
        unused: false
      },
      output: {
        'max_line_len': 1000000
      }
    }),
  ],
  debug: false,
  log: true
})
```

The `app.js` and `app.map.js` files are generated but my browser (Chrome) doesn't detect the source

![image](https://cloud.githubusercontent.com/assets/6224111/26167290/63878bac-3b04-11e7-87e3-ef9b3528a214.png)

nchanged:
We never handled it. It's possible but requires an update to the UglifyJS Plugin.

nchanged:
### FuseBox v4 Major Upgrade

As of today, the latest major version of FuseBox `v4` is available in `fuse-box@next`. FuseBox has undergone a full re-write from scratch, learning from its mistakes in the previous versions.

There are a lot of improvements, speed and config wise.

Get started [here](https://github.com/fuse-box/fuse-box/blob/master/docs/getting-started/get-started.md)

We are still working on the documentation and polishing before the major release. We are getting very close. And we need help testing.

```bash
npm install fuse-box@next
```

Most of the issues that I am closing have been resolved, however, if you find that your issue is still relevant in `v4` don't hesitate to re-open or (better) create a new one with `v4` configuration.

PR #787, opened 2017-09-08
Fix sass issues, default sourceRoot, adjust ts sourceMaps and prevent…

zigmo:
… vendor sourceMaps if not enabled. Fix for #778

### DIFF
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -40,7 +40,7 @@ export class BundleSource {
      * @type boolean
      * @memberOf BundleSource
      */
-    private includeSourceMaps: boolean = false;
+    public includeSourceMaps: boolean = false;
     
     public bundleInfoObject: any;
 
@@ -278,11 +278,6 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
      * @memberOf BundleSource
      */
     public getResult() {
-       
-        return {
-            content: this.concat.content,
-            sourceMap: this.includeSourceMaps ? this.concat.sourceMap : null,
-            add: this.concat.add
-        }
+        return this.concat;
     }
 }
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -281,7 +281,8 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
        
         return {
             content: this.concat.content,
-            sourceMap: this.includeSourceMaps ? this.concat.sourceMap : null
+            sourceMap: this.includeSourceMaps ? this.concat.sourceMap : null,
+            add: this.concat.add
         }
     }
 }
--- a/src/BundleSource.ts
+++ b/src/BundleSource.ts
@@ -32,6 +32,16 @@ export class BundleSource {
 
     private collectionSource: any;
 
+    /**
+     * Provied an info if this BundleSource should
+     * include the sourceMaps
+     *
+     * @private
+     * @type boolean
+     * @memberOf BundleSource
+     */
+    private includeSourceMaps: boolean = false;
+    
     public bundleInfoObject: any;
 
     /**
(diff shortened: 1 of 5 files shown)


src/plugins/stylesheet/SassPlugin.ts
```diff
@@ -60,6 +60,7 @@ export class SassPluginClass implements Plugin {
 
         const options = Object.assign({
             data: file.contents,
+            file: context.homeDir+"/"+file.info.fuseBoxPath,
```
nchanged: In this case we need to add it to Post and Less too! It's not just the only plugin.

src/core/File.ts
```diff
@@ -590,11 +592,7 @@ export class File {
             let jsonSourceMaps = JSON.parse(this.sourceMap);
             jsonSourceMaps.file = this.info.fuseBoxPath;
             jsonSourceMaps.sources = jsonSourceMaps.sources.map((source : string) => {
-                if(source.indexOf('stdin') !== -1) {
```
nchanged: Anything we can do about stdin? Why did you remove it?

src/BundleSource.ts
```diff
@@ -139,6 +149,11 @@ export class BundleSource {
             return;
         }
 
+        if(!this.includeSourceMaps) {
+            // set to true if this file has relevant sourceMaps
+            this.includeSourceMaps = file.belongsToProject() && this.context.sourceMapsProject || !file.belongsToProject() && this.context.sourceMapsVendor;
```
nchanged: 👍

src/plugins/stylesheet/SassPlugin.ts
zigmo: This need to be checked. Don't know if its required. This is a [special behaviour](https://github.com/sass/node-sass#special-behaviours) of node-sass. `file` is only used for sourceMaps if data is provided.

src/core/File.ts
zigmo: `stdin` shouldn't be there anymore because `file` is additionally provided to node-sass options - we will have the right file name with relativ path instead `stdin`. See SassPlugin comment.

zigmo:
Cool - maybe this can be a separated/expiremental pre-beta-release or so - so we can better test it. I think this would also solve this issue #543

nchanged:
Ok. it's here `fuse-box@2.3.1-test.1` @rs3d

rs3d:
Great job and thank you for the help @zigmo @nchanged!  The file structure for source files is working now.

**Some minor issues:**
1. `file":"index.scss"`should be named `index.css` to properly fit the generated filename, right?https://github.com/rs3d/fuse-box-sass-issue/blob/master/dist/index.scss.map#L1 
image



2.  Maybe a different issue: `outFile` without using `file` as filename: 
```
CSSPlugin({
         outFile: file => `${distPath}/module-XXX.css`
}) 
```
This generates `module-XXX.css` and in the same folder `module-XXX.css.map`.
But the resulting source map path is: 
`/*# sourceMappingURL=index.scss.map */` which should be 
`/*# sourceMappingURL=module-XXX.css.map */` to make this fully working.

nchanged:
@rs3d 
`outFile: file => `${distPath}/module-XXX.css`
that's dynamic

You should use it like so:

```
outFile: file => `${distPath}/${ file}.css
```

rs3d:
@nchanged Yes I know this isn't how it's intended. In my case it's a whitelabel platform in different languages, so my css files have a platform specific prefix. 

Maybe this is a more legit example?
```
outFile: file => `${distPath}/platform-${file}
```
Or for different languages (like arabic):
```
outFile: file => `${distPath}/ar-AE-${file}
```

rs3d:
Never mind, I'm switching to folder based structure now :) So i  don't need that many workarounds. Thx.

nchanged:
Cool!

PR #793, opened 2017-09-12
feat: add language target to control transpilation level

calebboyd:
Here is my approach, let me know what you think 

With this we can probably detect all sorts of language features and do the appropriate transpilation

I went with your suggestion and target now accepts a language target like: `target: 'browser@esnext'`


nchanged:
Terrific!
One suggestion, if you can add a traversal plugin here

https://github.com/fuse-box/fuse-box/blob/9d0ddd9509973e8bcb61a866efae816ae804b993/src/analysis/FileAnalysis.ts#L20

It would make sense to have a place where we could define levels automatically.

https://github.com/fuse-box/fuse-box/blob/9d0ddd9509973e8bcb61a866efae816ae804b993/src/analysis/FileAnalysis.ts#L149-L154

And I think we might have a buggy situation over here:

https://github.com/fuse-box/fuse-box/blob/9d0ddd9509973e8bcb61a866efae816ae804b993/src/analysis/FileAnalysis.ts#L158-L160

As `this.ast` is not overloaded

It is set over here for instance.

https://github.com/fuse-box/fuse-box/blob/9d0ddd9509973e8bcb61a866efae816ae804b993/src/analysis/plugins/ImportDeclaration.ts#L112-L121

To avoid that, we might need to parse the code again and set the correct AST


I am not sure, but I think we need to regenerate the code before transpiling it, IF `requiresRegenaration` is set

https://github.com/fuse-box/fuse-box/blob/9d0ddd9509973e8bcb61a866efae816ae804b993/src/analysis/plugins/ImportDeclaration.ts#L66

calebboyd:
Can you elaborate on what you mean by 

>It would make sense to have a place where we could define levels automatically.

calebboyd:
@nchanged I added https://github.com/fuse-box/fuse-box/pull/793/commits/a07b7f1d769d3a71f209a90b6321a1a163376e15

nchanged:
Awesome, I would like to push one commit, could you give me an access? 
Of course you can fix it yourself

```
if (this.requiresRegeneration) {
              this.file.contents = this.file.context.generateCode(this.ast);
          }
```

This will be useless if `requiresTranspilation` is set and all aliases will be gone...

calebboyd:
Not sure I fully understand -- Added access https://github.com/calebboyd/fuse-box/invitations

Issue #795, opened 2017-09-13
Requiring relative paths with package.json#main entries


calebboyd:
PathMaster#testFolder does not take into account a package.json being in the folder

https://github.com/fuse-box/fuse-box/blob/8134ac6d151daf98d409f09389562dc49ecc7e44/src/core/PathMaster.ts#L227-L248

It should probably perform a proper `require.resolve` resolution here. 

Do you have that utility extracted anywhere?

nchanged:
Yeah, that's a known bug, nope it's not there, I hope it won't affect the performance much

calebboyd:
I'd think something like `tryReadPkgMain(dir) : string | false` could perform pretty fast

nchanged:
It's actually here

https://github.com/fuse-box/fuse-box/blob/8134ac6d151daf98d409f09389562dc49ecc7e44/src/core/PathMaster.ts#L287-L301

calebboyd:
Here is the broken case:

```
|____folder 
| |____module.js //require('..')
|____library.js
|____package.json //main: './library.js'
```

nchanged:
It works only with explicit package require, like `module/hello/world`

calebboyd:
Right, I modified testFolder, to resolve the correct entry file -- but failed

calebboyd:
@devmondo Why the enhancement label? this is most certainly a bug

calebboyd:
Can you explain what `resolved` and `alias` are supposed to be?

calebboyd:
I made a gist that mostly demonstrates some not so fun issues...
https://gist.github.com/calebboyd/48d62efd1f36d137a9e0492837d81b18

nchanged:
@calebboyd alias is being replaced at runtime

nchanged:
@calebboyd check out this one, please
https://github.com/fuse-box/fuse-box/commit/cb5dfbd4670344b839fc81b9e2b7f969a6f45d85

I think removing `if explicit` fixes it, so it's checking it every time. Please verify, if it fixes the case, let's update the loader with that corner case you've found.

calebboyd:
Hey that looks good. I've updated my PR.
I also added a check so it doesn't resolve aliases to `'...js' or `../.js` `../../.js` etc..

nchanged:
Ok, let's see what's happening here `fuse-box@2.3.1-beta.7`
Let's test intensively

calebboyd:
Will do. Trying on large project rn

[...]

PR #797, opened 2017-09-14
fix: check pkg main

calebboyd:
This correctly identifies the main file (afaict)

But its still not resolving to the correct module. (null at runtime), so some tests are probably broken


nchanged:
Seems like a legit fix to me, let's figure out what's wrong with tests, we need to make them pass

calebboyd:
linking: #795

calebboyd:
So this fix doesn't work.. It works in the context of "`testFolder`" But the bundle is not resolving the correct file. If you can point me in the right direction I can continue to investigate, but Im not very familiar with this portion of fusebox

nchanged:
well it should be here
https://github.com/fuse-box/fuse-box/blob/91b4c47843eb11b4521af202f8875d1b6accfa65/src/core/PathMaster.ts#L233-L246

That's for sure, and has to return alias:

https://github.com/fuse-box/fuse-box/blob/91b4c47843eb11b4521af202f8875d1b6accfa65/src/core/PathMaster.ts#L296-L299

Actually, alias should come here:

https://github.com/fuse-box/fuse-box/blob/91b4c47843eb11b4521af202f8875d1b6accfa65/src/core/PathMaster.ts#L311-L313

calebboyd:
I'm out for the night. I think I've fixed the issue. Please review!

gulpfile.js
```diff
@@ -326,12 +326,8 @@ gulp.task("installDevDeps", function(done) {
         "buble",
     ];
 
-    if (os.platform().match(/^win/)) {
-        let windowsCommands = ["start", "cmd.exe", "/K", "npm", "install", "--no-save", ...deps];
```
nchanged: why did you remove that one?

src/loader/LoaderAPI.ts
```diff
@@ -293,9 +293,14 @@ function $getRef(name: string, o: RefOpts): IReference {
         wildcard = validPath;
     }
     if (!file && !wildcard) {
-        // try folder index.js
+        // try index.js
         validPath = $pathJoin(filePath, "/", "index.js");
         file = pkg.f[validPath];
+
+        if (!file && pkg.s && pkg.s.entry) {
```
nchanged: what's that? Why did you modify the loader?

gulpfile.js
calebboyd: I replaced it with `spawn('npm.cmd', ...)` on windows

This avoids a shell popping up, that has to be closed manually

src/loader/LoaderAPI.ts
calebboyd: If the path is fully relative it will land here. See:
https://github.com/calebboyd/fuse-test 

If its not the index it must be an entry file. Specifically, a fully relative name.

`'..'` or `'../../'`. As a guard an `isRelative` check could be made..

Thoughts?

nchanged: But what if there a package.json with a different entry? that's why we had that `alias` thingy in PathMaster, that overrides it

calebboyd: If you have an idea of what `alias` needs to be set to, any tips would be helpful.

calebboyd: From what I can tell providing an alias does nothing -- The property is set but then it is never  used again...
Is that correct?
https://github.com/fuse-box/fuse-box/blob/813bc793c7d9de764a5f9aac64889233426f2c0a/src/core/PathMaster.ts#L153-L155

src/core/PathMaster.ts
```diff
@@ -328,7 +328,7 @@ export class PathMaster {
                 let folder = path.isAbsolute(name) ? name : path.join(root, name);
                 const folderPath = this.testFolder(folder, name);
                 if (folderPath) {
-                    return { resolved: folderPath }
+                    return { resolved: folderPath, alias: name }
```
calebboyd: Adding alias here doesn't really do anything

src/core/PathMaster.ts
```diff
@@ -224,33 +226,49 @@ export class PathMaster {
...
-                    let result = path.join(name, index);
-                    let startsWithDot = result[0] === "."; // After transformation we need to bring the dot back
-                    if (startsWithDot) {
-                        result = `./${result}`;
-                    }
-                    return result;
+            let result
+            const pkg = this.tryReadPkg(folder)
+            if (pkg && pkg.main) {
+                result = path.join(name, pkg.main)
```
nchanged: This should return an alias (FuseBox will regenerate the code and point to a correct file)

src/loader/LoaderAPI.ts
```diff
@@ -293,9 +293,13 @@ function $getRef(name: string, o: RefOpts): IReference {
         wildcard = validPath;
     }
     if (!file && !wildcard) {
-        // try folder index.js
+        // try index.js
         validPath = $pathJoin(filePath, "/", "index.js");
         file = pkg.f[validPath];
+
+        if (!file && filePath === ".") {
```
nchanged: Do we still need to modify the loader? Which case does it cover? `require(".")`?
Shouldn't it require index if exists?

nchanged:
Amazing, almost there, let's verify a few things before merging

src/loader/LoaderAPI.ts
calebboyd: Yeah, So here is an example of it being hit
https://github.com/calebboyd/fuse-test

Adding a log in that if block (L301)
`console.log('loading: "', name, '" filePath: "', filePath, '" entry: "', pkg.s.entry, '"')`

-> 
```
loading: " ../ " filePath: " . " entry: " test.js "
loading: " ../../ " filePath: " . " entry: " test.js "
```

calebboyd: It covers any fully relative path that resolves to the package.json in the pkg context

dgreif:
Bundled a new project with fusebox today.  It uses `request-promise`, which does some interesting require paths in `require-promise-core`.  It failed with `2.2.31`, but worked with `2.3.1-beta.14`.  This is a real-world case where @calebboyd's fix is needed.

PR #798, opened 2017-09-14
Postcss fix: pass options through to plugins

### LINKED ISSUE #788: PostCSSPlugin problems
In the PostCSSPlugin,
we pass processors and options.
In the documentation, the options.plugins "seems" to replace processors instead of it should work like options if we refer to the postCSS JS api...

**PostCSSPlugin.ts**
```
return postcss(this.processors)
            .process(file.contents, postCSSPlugins)
            .then(result => {
                file.contents = result.css;

                if (file.context.useCache) {
                    file.analysis.dependencies = cssDependencies;
                    file.context.cache.writeStaticCache(file, generateSourceMaps && file.sourceMap, "postcss");
                    file.analysis.dependencies = [];
                }
                return result.css;
            });
```
**postCSS api**
```
postcss([processors])
            .process(css, {options})
```

Furthemore,
we can't pass options to postcss...

PS : To make it works for me and only me, I replaced **postCSSPlugins** by **this.options**
Maybe I'm the only one to use postcss ^^

geowarin:
Should fix #788.

Sorry if I highjacked @ben33's pull request but we really need this to migrate to the new fuse-box version.
Please @ben33, tell me if it also solves your problem! Your review is very welcome.

In this PR, I migrated the old postcss tests.
I also added a new one, highlighting the issue mentioned in #788.

Some plugins, like the [url-plugin](https://github.com/postcss/postcss-url), need the postcss options to be passed through to work correctly.

---

@nchanged, I also think fusebox should follow some of [the guidelines](https://github.com/postcss/postcss/blob/master/docs/guidelines/runner.md) provided by the postcss team.

Especially for the warning and the error processing.
I was thinking about adding a `.catch()` function to handle this in the plugin code, a bit like the less plugin does.

Is it a good idea?

### DIFF
--- a/docs/plugins/css/PostCSSPlugin.md
+++ b/docs/plugins/css/PostCSSPlugin.md
@@ -96,7 +96,7 @@ Or like this:
 ```js
 FuseBox.init({
     plugins : [
-         [PostCSSPlugin([], {
+         [PostCSSPlugin({
              // postcss plugins
              plugins: [require('postcss-url')({url: "rebase"})],
              // should fusebox generate sourcemaps (see below), default: true
--- a/docs/plugins/css/PostCSSPlugin.md
+++ b/docs/plugins/css/PostCSSPlugin.md
@@ -75,6 +75,27 @@ If you require more options, you can pass them to the plugin:
 ```js
 FuseBox.init({
     plugins : [
+         [PostCSSPlugin(
+             // postcss plugins
+             [require('postcss-url')({url: "rebase"})], 
+             {
+                 // should fusebox generate sourcemaps (see below), default: true
+                 sourceMaps: false,
+                 // additional paths for css resolution (see below)
+                 paths: [],
+                 // all other options will go to the postcss process function (see below) 
+                 parser: parser,
+            }
+         ), CSSPlugin()]
+    ]
+});
+```
+
+Or like this
+
+```js
+FuseBox.init({
+    plugins : [
          [PostCSSPlugin({
              // postcss plugins
              plugins: [require('postcss-url')({url: "rebase"})],
--- a/docs/plugins/css/PostCSSPlugin.md
+++ b/docs/plugins/css/PostCSSPlugin.md
@@ -1,7 +1,12 @@
 # PostCSS Plugin
 
 ## Description
-Allows using PostCSS, A tool for transforming CSS with JavaScript.
+
+Allows using PostCSS, a tool for transforming styles with JS plugins. 
+These plugins can lint your CSS, support variables and mixins, 
+transpile future CSS syntax, inline images, and more.
+
+Check [PostCSS website](http://postcss.org/) for more information.
 
 ## Install
 
@@ -12,12 +17,12 @@ npm install postcss --save-dev
 ```
 
 ## Usage
-check [PostCSS website](http://postcss.org/) for more information.
-note: The PostCSS plugin generates CSS, Therefor it must be chained prior to the CSSPlugin to be used.
+
+The PostCSS plugin generates CSS. It must therefore be chained prior to the CSSPlugin to be used.
 
 ### Setup
 
-Import from FuseBox
+Import the plugin from FuseBox
 
 ```js
 const {PostCSSPlugin} = require("fuse-box");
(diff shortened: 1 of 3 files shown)


geowarin:
Other considerations:

* I think we should remove the `plugin` attribute from the options, I don't really know what to do with this.
I don't think it is a breaking change because it is unclear what it was supposed to do.

* I don't really understand the `path` option but I suppose we could use postCss `from` option instead? What do you think @nchanged ?

* We should update the documentation

src/plugins/stylesheet/PostCSSPlugin.ts
```diff
@@ -58,34 +58,29 @@ export class PostCSSPluginClass implements Plugin {
...
-        paths.push(file.info.absDir);
+        let generateSourceMaps = true;
+        let postCssOptions;
+        let pathOptions: string[] = [file.info.absDir];
+
+        if (this.options) {
+            const {sourceMaps = generateSourceMaps, plugins, paths = pathOptions, ...otherOptions} = this.options;
+            generateSourceMaps = sourceMaps;
+            postCssOptions = otherOptions;
+            pathOptions = paths;
```
nchanged: I think `pathOptions` should always contain file.info.absDir no matter what, otherwise CSSDependencyExtractor will not work. Or am I missing something?

src/plugins/stylesheet/PostCSSPlugin.ts
```diff
@@ -58,34 +58,29 @@ export class PostCSSPluginClass implements Plugin {
...
         if (!postcss) {
             postcss = require("postcss");
         }
-        let generateSourceMaps = true;
-        let postCSSPlugins = [];
-        if (Array.isArray(this.options)) {
-            postCSSPlugins = this.options;
-        } else {
-            if (this.options) {
-                if (Array.isArray(this.options.plugins)) {
```
nchanged: Can we still keep it compatible with the old version? I believe some people are relying on it. Let's just update the doc to the latest

src/tests/plugins/PostcssPlugin.test.ts
```diff
@@ -0,0 +1,129 @@
+import {createEnv} from "../stubs/TestEnvironment";
+import {should} from "fuse-test-runner";
+import {CSSPlugin} from "../../plugins/stylesheet/CSSplugin";
+import {PostCSS} from "../../plugins/stylesheet/PostCSSPlugin";
+
+const postcss = require("postcss");
+
+const changeDisplayPlugin = postcss.plugin("ChangeDisplayPlugin", function (options) {
```
nchanged: 👍

nchanged:
@geowarin we need path along with the initial `file.info.absDir` for the CSSDependencyExtractor

src/plugins/stylesheet/PostCSSPlugin.ts
geowarin: You're not missing something, I set it up so it only contains `file.info.absDir` only if `paths` is not defined.
I'll fix this.

src/plugins/stylesheet/PostCSSPlugin.ts
geowarin: Ok but the API will look weird:

There is two options to declare postcss plugins:
```js
[PostCSS(plugins)]
```

and : 

```js
[PostCSS([], {plugins: plugins})]
```

What happens if both are defined? Should we merge them?

nchanged: I think it was supposed to look like:

```
[PostCSS({plugins : []})]
```
Like here http://fuse-box.org/plugins/postcss-plugin#paths-for-hmr

nchanged: I think it's either the proper way (you suggested) or the old one, but we need to clean up the docs

geowarin: Ok I set it up to merge them, this ensures maximum compatibility right now.
This way, we can defer this decision if you want.

I like you suggestion but it would be a breaking change.
Your call 😄

nchanged: Why a breaking change? That's how it is working now
either
`[PostCSS({plugins : []})]` or `[PostCSS([]]`

WE just need to check for the first argument and check if it's an object with `plugins` property how it was before

geowarin: Unfortunately, I think it is broken right now.
According to the postcss docs, we can only pass the plugins (what we refer in the code as processors) as the first argument of the postcss function.

What fusebox [does today](https://github.com/fuse-box/fuse-box/blob/8134ac6d151daf98d409f09389562dc49ecc7e44/src/plugins/stylesheet/PostCSSPlugin.ts#L88) does not work.

As you can see the [processOptions](http://api.postcss.org/global.html#processOptions) (second argument to `postcss().process()`) does not take plugins.

But I'll make it work as you suggested, no problem.

nchanged: Oh, that's a bummer ( Can you update the docs please?

nchanged: Hm, i think you got me wrong here ;-( I liked your suggestion - to align with postcss style, as in
```js
PostCSS([plugin1, plugin2], {  /options/ })
```

And I think it should be like that... It's just wanted to keep the old style and update the documentation (removing the docs on the incorrect one).

To avoid breaking changes, you know?

geowarin: No problem @nchanged 😃 
I'll go back to concatenating both "processors" and the plugins option then, right ?

nchanged: like in PostCSS yes, except make an additional condition for the first argument, if it's an object, take `plugins` and the pass the rest to postcss options

geowarin:
Ok so I hope I got it right this time.

The postcss plugin takes either:

1. An array of plugins and an optional option argument
2. An option argument where you can also add plugins.

In case we get both plugin from the first argument and the second, we concatenate them

docs/plugins/css/PostCSSPlugin.md
```diff
@@ -42,58 +47,129 @@ FuseBox.init({
...
+```
+
+Or like this:
+
+```js
+FuseBox.init({
+    plugins : [
+         [PostCSSPlugin({
+             // postcss plugins
+             plugins: [require('postcss-url')({url: "rebase"})],
```
geowarin: @nchanged Do you want to remove this to discourage the usage of this option ?

geowarin:
If we want to get rid of the plugin option in the mid-term, I could do

```patch
Index: src/tests/plugins/PostcssPlugin.test.ts
===================================================================
--- src/tests/plugins/PostcssPlugin.test.ts	(revision bb25dfeb9fd2d1ea86c4575202e83e49edfb1d1b)
+++ src/tests/plugins/PostcssPlugin.test.ts	(revision )
@@ -77,7 +77,7 @@
   "Should execute several plugins"() {
     return setup({
       plugins: [
-        [PostCSS([changeDisplayPlugin], {plugins: [idToClassPlugin]}), CSSPlugin({})]
+        [PostCSS([changeDisplayPlugin, idToClassPlugin]), CSSPlugin({})]
       ]
     })
       .then((result) => {
Index: src/plugins/stylesheet/PostCSSPlugin.ts
<+>UTF-8
===================================================================
--- src/plugins/stylesheet/PostCSSPlugin.ts	(revision bb25dfeb9fd2d1ea86c4575202e83e49edfb1d1b)
+++ src/plugins/stylesheet/PostCSSPlugin.ts	(revision )
@@ -6,7 +6,6 @@
     [key: string]: any;
     paths?: string[],
     sourceMaps?: boolean;
-    plugins?: any[]
 }
 
 export type Processors = (() => any)[];
@@ -60,7 +59,6 @@
 
         const {
             sourceMaps = true,
-            plugins = [],
             paths = [],
             ...postCssOptions
         } = this.options || {};
@@ -77,7 +75,7 @@
             postcss = require("postcss");
         }
 
-        return postcss(this.processors.concat(plugins))
+        return postcss(this.processors)
             .process(file.contents, postCssOptions)
             .then(result => {
                 file.contents = result.css;
@@ -98,7 +96,10 @@
     if (Array.isArray(processors)) {
         return new PostCSSPluginClass(processors, opts);
     }
-    return new PostCSSPluginClass([], processors);
+    // We still take plugins from options for legacy reasons
+    // It is discouraged and does not appear in type definitions
+    const {plugins = [], ...otherOptions} = processors || {};
+    return new PostCSSPluginClass(plugins, otherOptions);
 }
 
 export {PostCSS}
```

and remove it from the docs

docs/plugins/css/PostCSSPlugin.md
nchanged: Let's keep it please, we will remove it in the next release. Probably deprecation warning?

geowarin:
@nchanged Anything else I can do to ?

nchanged:
Awesome!!!

geowarin:
Hey @nchanged ! Any plan to release this ?

nchanged:
@geowarin very soon. You can just install the latest beta

geowarin:
Oh indeed, I forgot about beta. Thanks!

PR #802, opened 2017-09-16
fix: quantum relative package requires

calebboyd:
That wasn't too bad. Let me know what you think.

Quantum is pretty neat. Great Work!


nchanged:
Looks awesome! Thanks a lot! Love your coding style

PR #822, opened 2017-09-24
Fix out of project relative imports

BusinessDuck:
The bug regarding with relative paths, when u try to import files from outside of project dir.

One 
Two 
Three


nchanged:
You can't get out of projects' directory. That's the rule, and it cannot be fixed, as it's not a bug.

PR #829, opened 2017-09-28
globalRequire in case of npm (server)

eddow:
Finalising the `globalRequire` idea :
The flag really solves a lot of issues. The only thing here is that it was not taken into consideration in the `npm` or `server` target case.

With this PR, it affects all the cases of `global.require` references.

### DIFF
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -50,9 +50,11 @@
 
 
     /* @if server */
+    /* @if globalRequire */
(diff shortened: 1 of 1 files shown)


nchanged:
https://github.com/fuse-box/fuse-box/blob/467d7f5dc37e5161c9881d37c496528279dddeaa/modules/fuse-box-responsive-api/index.js#L13-L16

Isn't it there already?

nchanged:
Oh yeah, you right.

PR #830, opened 2017-09-28
fix: respect log level for cache clear

calebboyd:


src/Log.ts
```diff
@@ -97,9 +101,15 @@ export class Log {
...
             return null;
-        });;
-    }
+        });
 
+        setTimeout(() => {
+            if (this.printLog) {
+                Log.deferred.forEach(x => x(this))
+            }
+            Log.deferred.length = 0
```
nchanged: Isn't it better to reset the array?

` Log.deferred = []`

calebboyd: It has the same effect, I can change it: changed

Issue #821, opened 2017-09-24
Babel's config in VuePlugin


s4kr4:
Hello,

In the `VuePlugin`, we can specify the transpiler by defining the `lang` attribute.
And if we select `babel`, it seems that we can set babel's config like this.

```javascript
VuePlugin({
  babel: {
    config: {
      "presets": [
        ["env", {
          "targets": {
            "browsers": [
              "last 2 versions",
            ],
          },
        }],
      ],
    },
  },
}),
```

But in this case, `babel` loses default config(`babel-plugin-transform-es2015-modules-commonjs`).
Is this behavior to be expected?
I think that specified settings should be merged with default settings.

Thanks.

devmondo:
@s4kr4  we are working with Mr. @Shepless on making the vue plugin perfect, hang in there, it will be out soon :)

nchanged:
Yeah there will be a fully fledged support for Vue in 2.3.1 it’s coming in 1 week

s4kr4:
Sounds good!

TombolaShepless:
@s4kr4 hey there - I really want a fully functional one too for my own work too. Im working on the plugin now and it is almost done but I am working on it in the evenings in personal time so please bear with me 👍

TombolaShepless:
Update for you @s4kr4 Functionality is developed - its not perfect but it is a lot more feature complete now. I need to write the documentation and will have a PR up tomorrow evening for @nchanged and @devmondo to review.

Shepless:
@s4kr4 can you verify that this issue is now resolved?

s4kr4:
@Shepless It's nice idea that using BabelPlugin instead of the VuePlugin's inner process. This resolved my this issue related to Babel's config.
This issue will close, but now I have another issue on VueComponentPlugin.
I'll make new issue about that.

Thanks.

Shepless:
@s4kr4 good news. Can't see your new issue? Did you manage to resolve?

s4kr4:
@Shepless Sorry for late, I made #835 .

Issue #542, opened 2017-05-17
Vue import are not bundled


eddow:
When a `file.vue` imports a module (local or npm), this one is not included in the result bundle, resulting in an `Error: Cannot find module`

https://github.com/eddow/eddow-bug-expo/tree/vue-import

devmondo:
it is `import * as testImp from './testImp'` not `import * as testImp from 'testImp'` as it is relative file and not node module.

will close this for now, feel free to open otherwise :)

eddow:
The typo is corrected - the bug is still there

nchanged:
@eddow you think the latest implementation fixes it? 
https://github.com/fuse-box/fuse-box/blob/master/src/plugins/VuePlugin.ts#L59

eddow:
Is this change already deployed in 2.2.0?
I just tried with the little bug-expo after upgrading fuse-box, it didn't made a change there.

nchanged:
have you tried with the latest beta?

eddow:
Okay, with the beta.12, I first had to solve `Error: Cannot find module 'acorn'`.
Once solved, the test unfortunately let the results unchanged.
I had a peek inside of the last modifications of VuePlugin.js and we effectively have the last version in beta.12. Looking around, I saw no modification that might have corrected it so far indeed.

eddow:
So, as the bug is still not resolved, I investigated a bit (I for now have to maintain a library files as importing files from a `.vue` source does not make the import bundled)

In [File.consume](https://github.com/fuse-box/fuse-box/blob/4e519489694f00a8b25d0aab838267d480ca6ea3/src/core/File.ts#L387), we can see that `.ts`, `.js` and other files have different treatment. "Other" files have a treatment through the plugin - and this treatment occurs in a promise that is [queued](https://github.com/fuse-box/fuse-box/blob/4e519489694f00a8b25d0aab838267d480ca6ea3/src/core/File.ts#L265)
Therefore, the dependencies of a `.vue` file, that will be used [right after](https://github.com/fuse-box/fuse-box/blob/4e519489694f00a8b25d0aab838267d480ca6ea3/src/core/ModuleCollection.ts#L383) the file is `consume`-d will be empty as this will happen before the promise is effected, the plugin applied and the `.vue` file analysed

nchanged:
Are we still seeing it?

eddow:
I tried with the last version from git and it is solved !
[...]

PR #832, opened 2017-09-29
Better Vue File Support

TombolaShepless:
This PR adds the new plugin `VueComponentPlugin`:

#### Features that the old `VuePlugin` does not contain:
- Adds in transpilation/pre-proessing for all blocks (`<template>`, `<script>`, `<style>`)
- Adds scoped styling functionality
- Adds ability to load external files using `src` attribute
- Has a default plugin chain for each block, but also allows per file overriding using `lang` attribute, or, global override using plugin chains
- HMR integration

I've added a deprecation warning in the old `VuePlugin` so users know to migrate over as I assume we will be removing the old plugin in due course.

Massive thanks to @nchanged and @devmondo for their help in building this plugin!

UPDATE: This will close the following: #125 #542 #821


TombolaShepless:
Sure, I'll add it back in with a big console deprecation message :) gimme 5 mins.

package.json
```diff
@@ -99,11 +99,13 @@
     "mustache": "^2.3.0",
     "opencollective": "^1.0.3",
     "postcss": "^6.0.1",
+    "postcss-selector-parser": "^2.2.3",
     "pretty-time": "^0.2.0",
     "prettysize": "0.0.3",
     "realm-utils": "^1.0.8",
     "request": "^2.79.0",
     "shorthash": "0.0.2",
+    "vue-hot-reload-api": "^2.1.0",
```
nchanged: We can't put it here, let's update the doc what's required to be installed.

src/index.ts
```diff
@@ -12,7 +11,7 @@ export { IQuantumExtensionParams } from "./quantum/plugin/QuantumOptions";
 export { ComputedStatementRule } from "./quantum/plugin/ComputerStatementRule";
 export { QuantumPlugin } from "./quantum/plugin/QuantumPlugin";
 export { ReplacePlugin } from "./plugins/ReplacePlugin";
-export { VuePlugin } from "./plugins/VuePlugin";
```
nchanged: Let's keep the old plugin! We can remove it after a few releases!

src/plugins/vue/VuePlugin.ts
```diff
@@ -0,0 +1,266 @@
...
+import { CSSPluginClass } from "../stylesheet/CSSplugin";
+import { Concat, hashString } from "../../Utils";
+import { VueBlockFile } from './VueBlockFile';
+import { VueTemplateFile } from './VueTemplateFile';
+import { VueStyleFile } from './VueStyleFile';
+import { VueScriptFile } from './VueScriptFile';
+import * as path from "path";
+import * as fs from "fs";
+import { each } from "realm-utils";
+const vueCompiler = require("vue-template-compiler");
```
nchanged: Should be hidden with the plugin. Can't be in dependencies.

TombolaShepless: As in:

```js
transform(file) {
 const vueCompiler = require('vue-template-compiler')
}
```

Guess I'd need to update docs for dependency listing too

nchanged: That's right! The installation must include all dependencies

src/plugins/vue/VuePlugin.ts
```diff
@@ -0,0 +1,266 @@
...
+    cacheItem[path] = {
+      contents,
+      sourceMap
+    };
+  }
+
+  public bundleEnd(context: WorkFlowContext) {
+    if (context.useCache) {
+      context.source.addContent(`
+        var process = FuseBox.import('process');
```
nchanged: Let's make sure process is included for sure. Put it into file dependencies.
And I would suggest wrapping it
```
if( process.env.NODE_ENV !== "production) {
 // that code
}
```

TombolaShepless: Updated.

eddow:
Woaw; I might have many projects to migrate but I will do it with pleasure, I'm looking forward to have time for this!
Note: it might mean "bug reports" :-p

PR #844, opened 2017-10-03
Commit Message Validation & Formatting

Shepless:
Feel free to close, but please pull this branch down and have a play to see what you think:

- `npm i`

Then do a git commit. Expected message format is `type(scope): short message`, e.g. `feat(bundler): Made the bundler do something cool`. If it doesnt follow this format then it will error.

Let me know what your thoughts are.

### UPDATE
The latest version of this PR will accept either of the following commit messages as valid:

- `type(scope): short message`
- `type: short message`

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,9 +121,9 @@
   "config": {
     "validate-commit-msg": {
       "scope": {
-        "required": true,
+        "required": false,
         "allowed": ["*"],
-        "validate": true,
+        "validate": false,
         "multiple": false
       },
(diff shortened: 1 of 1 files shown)


Shepless:
@nchanged as discussed in gitter - I will play with the message format but I will have to do it tomorrow evening.

Shepless:
@nchanged I've updated so the scope is no longer validated and can be omitted. Let me know what you think.

Shepless:
@nchanged @calebboyd you guys get a chance to play with the updated rules?

nchanged:
Can we put example commit messages in the description?

Shepless:
Updated description

PR #851, opened 2017-10-05
VueComponentPlugin - Dependency Fixes

Shepless:
Moved TypeScript and PostCSS imports out of module scope as it was causing issues for people not using these modules in their projects.


src/plugins/vue/VueStyleFile.ts
```diff
@@ -19,6 +18,8 @@ export class VueStyleFile extends VueBlockFile {
   }
 
   private async applyScopeIdToStyles(scopeId: string) {
+    const postcss = require('postcss');
```
nchanged: postcss is in deps, but it's okay ;-)

Shepless: Aye just to be double sure ;)

PR #855, opened 2017-10-08
Cache Buster - Handle "Non-Standard" Runtime Environments

Shepless:
I'm trying to run FuseBox in places like AWS Lambda, Firebase Cloud Functions etc. but they all fail because they have read only file systems and its this check that is causing it.

You might be wondering why I am doing this? Its for a cool new tool for the FuseBox community but its nowhere near ready to share...

### DIFF
--- a/src/CacheBreaker.ts
+++ b/src/CacheBreaker.ts
@@ -4,19 +4,24 @@ import { fastHash, removeFolder, ensureDir } from "./Utils";
 import * as fs from "fs";
 import { Log } from "./Log";
 /**
- * 
+ *
  * The only purpose of this ugly function is to break cache when user changes the config..
- * 
+ *
  * @export
  */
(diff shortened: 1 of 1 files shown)


src/CacheBreaker.ts
```diff
@@ -28,4 +33,4 @@ export function breakCache() {
     } else {
         fs.writeFileSync(infoFile, fileModificationTime.toString());
```
nchanged: Shouldn't that be wrapped too?

Shepless: No - it won't ever reach here if the file system is read only. `ensureDir` checks if it exists (which it wont) then tries to write so it will always error before here.

nchanged: Oh right, didn't see the return statement!

PR #859, opened 2017-10-10
feat: Add ability to override file extension resolution

Shepless:
FuseBox can now be configured to look for customised extensions ahead of standard ones. This is
useful for platform specific builds. For example:

```js
FuseBox.init({
   extensionOverrides: ['.desktop.js']
});
```

With the above configuration FuseBox will favour loading contents from files with a `.desktop.js` extension over `.js` extension. If `.desktop.js` does not exist then normal resolution will occur. This a essentially a port of the Webpack functionality for [resolve extensions](https://webpack.js.org/configuration/resolve/#resolve-extensions).

In Summary:
- Added Functionality
- Added Unit Tests
- Added Documentation


src/core/ExtensionOverrides.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public add(override: string) {
+    if (this.isValid(override)) {
+      this.overrides.push(override);
+    }
+  }
+
+  public setOverrideFileInfo(file: File): string {
+    if (this.overrides.length === 0) {
+      return null;
```
nchanged: nitpick: better to return `undefined` (nothing)

src/core/File.ts
```diff
@@ -159,10 +161,14 @@ export class File {
...
             this.params = info.params;
         }
+
         this.absPath = info.absPath;
+
         if (this.absPath) {
             this.relativePath = ensureFuseBoxPath(path.relative(this.context.appRoot, this.absPath));
         }
+
+        this.context.extensionOverrides.setOverrideFileInfo(this);
```
nchanged: What is it doing in alternative content? Shouldn't it be at `consume` level? That one limits many plugins

src/core/ExtensionOverrides.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    const fileInfo = path.parse(file.info.absPath);
+
+    for (let overrideExtension of this.overrides) {
+      const overridePath = path.resolve(fileInfo.dir, `${fileInfo.name}${overrideExtension}`);
+
+      if (overrideExtension.indexOf(fileInfo.ext) > -1 && fs.existsSync(overridePath)) {
+        file.absPath = file.info.absPath = overridePath;
```
nchanged: i don't think we should override absPath ( my bad,  I told you that )
we should map it in the context instead for the watcher

src/tests/BundleApi.test.ts
```diff
@@ -135,6 +135,34 @@ export class BundleApiTest {
...
+    "Should not share extension overrides across bundles"() {
+        const fuse = createFuse();
+        const bundle1 = fuse.bundle("app")
+            .extensionOverrides(".foo.js", ".foo.json");
+        const bundle2 = fuse.bundle("app")
+            .extensionOverrides(".bar.js", ".bar.json");
+
+        should(bundle1.context.extensionOverrides.overrides).deepEqual([".foo.js", ".foo.json"]);
+        should(bundle2.context.extensionOverrides.overrides).deepEqual([".bar.js", ".bar.json"]);
+    }
```
nchanged: Not enough tests, we should make similar tests additionally

https://github.com/fuse-box/fuse-box/blob/8586d90baac5326f849aadf8aca82377a7e33764/src/tests/plugins/CSSPlugin.test.ts#L131-L147

To cover all the cases, not just VuePlugin

src/core/ExtensionOverrides.ts
nchanged: Using HMR_FILE_REQUIRED
https://github.com/fuse-box/fuse-box/blob/8586d90baac5326f849aadf8aca82377a7e33764/src/plugins/stylesheet/CSSplugin.ts#L132-L137

TombolaShepless:
@nchanged cheers for the feedback - will go through it tonight and update!

src/core/ExtensionOverrides.ts
nchanged: Also let's add a condition if it belongs to the actual project (vendors ignored)
`file.belongsToProject()`

src/core/File.ts
Shepless: It doesn't just do this in alternative content. We do this work in the `File` constructor so it happens at the lowest level possible. All testing I have done doesn't look to have impacted.

src/tests/BundleApi.test.ts
Shepless: Im not sure I understand? The tests are in the bundle? I should probably add some for `File` though for sure.

src/core/File.ts
nchanged: Then why is it happening on `addAlternativeContent` ?

src/tests/BundleApi.test.ts
nchanged: No, I just gave you an example. You can create test bundle in a test, with the actual file structure and stuff.

src/core/File.ts
Shepless: Its not?
screen shot 2017-10-16 at 20 43 47

nchanged: sorry man, i am tired ;-))

Shepless: hehe no worries bud.

Shepless:
Closing as the rebase went horribly wrong haha! I've sorted on a new branch.

PR #862, opened 2017-10-12
chore: remove unnecessary acorn plugins

calebboyd:
Updates acorn and removes invalidated plugins.

I'm not able to update `bin.js` which will cause the tests to fail. (manually updating it passes the tests)

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -2072,7 +2072,6 @@ ${file.headerContent ? file.headerContent.join("\n") : ""}`);
             const HeaderImport_1 = require("./HeaderImport");
             const acorn = require("acorn");
             const escodegen = require("escodegen");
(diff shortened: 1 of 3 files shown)


nchanged:
I don't think we should update bin.js
There must be some other reason why the tests fail...

calebboyd:
@nchanged That isn't the case, Take a look at my second commit.
The `make-test-runner` github task isn't working... 😕

nchanged:
Oh brilliant!
Yeah, that's some old version of FuseBox, i don't really want to update it ;-)
So are you sure that the new version of acorn doesn't require plugins? It's all safe this way?

calebboyd:
It should be! Probably need to let it marinate in a beta for a bit

PR #865, opened 2017-10-15
fix quantum code splitting issue

leon-andria:
- isBrowser variable is always available 
- code splitting config always take the browser cofnig, never the server


nchanged:
* IsBrowser should be available only for target browser
* Could splitting config should be in both bundles regardless, many people do universal apps and server uses code splitting too 
* You broke the tests badly.

PR #867, opened 2017-10-16
Update index.js

leon-andria:
- use correct split config options with quantum

### DIFF
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -235,7 +235,17 @@
             /* @if codeSplitting */
             if (bMapping.i && bMapping.i[id]) {
                 var data = bMapping.i[id];
-                req(bMapping.c.b + data[0], function(err, result) {
+                /* @if universal */
+                const contextualPath = isBrowser ? bMapping.c.b : bMapping.c.s;
+                /* @end */
+                /* @if server */
+                const contextualPath = bMapping.c.s;
+                /* @end */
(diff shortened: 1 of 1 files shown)


modules/fuse-box-responsive-api/index.js
```diff
@@ -235,7 +235,17 @@
             /* @if codeSplitting */
             if (bMapping.i && bMapping.i[id]) {
                 var data = bMapping.i[id];
-                req(bMapping.c.b + data[0], function(err, result) {
+                /* @if universal */
+                const contextualPath = isBrowser ? bMapping.c.b : bMapping.c.s;
```
nchanged: LGTM except for `const` please, convert it to `var` and I think we can merge it.
Good job!

nchanged: Oh there is one thing. `bMapping.c.b` can it be undefined ?! Can you check that please

leon-andria:
It cant be undefined its a config that you auto generate with default value. The moment you split, it's there. That's make sense.

nchanged:
Ok, then tend to my other commit, with replacing const to var

leon-andria:
The changes was done.

nchanged:
Lemme dwell on it a bit, i will merge it today

nchanged:
Why isn't the commit in the format of (type): message? 
Did you do npm install beforehand? it shouldn't have allowed you to commit like this

PR #870, opened 2017-10-16
feat: Allow typescript decorators for vue components

Shepless:
This allows vue components to work with decorators such as `vue-class-component`:

```js
import Vue from 'vue';
import Component from 'vue-class-component';

@Component({})
export default class VueClassComponent extends Vue {
   message: string = 'Hello!'
   onClick (): void {
      window.alert(this.message)
   }
}
```

All the user needs to do is install `tslib` and update their `tsconfig.json` to `importHelpers`:

```json
{
  "compilerOptions": {
    "importHelpers": true
  }
}
```

This also fixes a couple of other bugs:

- Hot reloading did not work for non-scoped styled components
- `DEFAULT_OPTIONS` were being extended each time a `VueComponentPlugin` instance was created.

### DIFF
--- a/src/tests/plugins/VueComponentPlugin.test.ts
+++ b/src/tests/plugins/VueComponentPlugin.test.ts
@@ -1,4 +1,4 @@
-import { VueComponentPlugin, BabelPlugin, HTMLPlugin, SassPlugin, TypeScriptHelpers } from "../../index";
+import { VueComponentPlugin, BabelPlugin, HTMLPlugin, SassPlugin } from "../../index";
 import { createEnv } from "../stubs/TestEnvironment";
 import { should } from "fuse-test-runner";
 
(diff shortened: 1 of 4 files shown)


src/tests/plugins/VueComponentPlugin.test.ts
```diff
@@ -1,4 +1,4 @@
-import { VueComponentPlugin, BabelPlugin, HTMLPlugin, SassPlugin } from "../../index";
+import { VueComponentPlugin, BabelPlugin, HTMLPlugin, SassPlugin, TypeScriptHelpers } from "../../index";
```
nchanged: `TypeScriptHelpers` should be probably removed

nchanged:
Shall we add this to the doc? Regarding `tslib`

PR #884, opened 2017-10-19
feat: Uglify-es support for quantum

bsl-zcs:
I asked at gitter for the best way to design an API for it, but got no reply.  So here is the simplest implementation possible, though it probably isn't very straightforward to use.

It picks up uglify's options from the `UglifyESPlugin` or could be triggered by quantum option `uglify: { es6: true }`

### DIFF
--- a/src/quantum/plugin/QuantumPlugin.ts
+++ b/src/quantum/plugin/QuantumPlugin.ts
@@ -24,6 +24,11 @@ export class QuantumPluginClass implements Plugin {
                     // remove uglify js
                     delete plugins[index];
                 }
+                if (plugin.constructor.name === "UglifyESPluginClass") {
+                    this.coreOpts.uglify = { es6:true, ...plugin.options };
+                    // remove uglify es
+                    delete plugins[index];
+                }
 
                 if (plugin.constructor.name === "WebIndexPluginClass") {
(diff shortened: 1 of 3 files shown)


src/quantum/plugin/QuantumPlugin.ts
```diff
@@ -24,6 +24,11 @@ export class QuantumPluginClass implements Plugin {
                     // remove uglify js
                     delete plugins[index];
                 }
+                if (plugin.constructor.name === "UglifyESPluginClass") {
+                    this.coreOpts.uglify = { es6:true, ...plugin.options };
```
nchanged: As far as I remember UglifyEs5 will complain about an unknown option.  When passing the option, let's make sure `es6 :true` is not there (delete it)

It would great if you added documentation for that here:
https://github.com/fuse-box/fuse-box/blob/master/docs/quantum.md#installation

Good job!

bsl-zcs: It got extracted later on via destructuring here: `{ es6 = false, ...opts }`

nchanged: Ok, can you add the docs please?

bsl-zcs:
Well, I hope it's comprehensible.  :)

Issue #885, opened 2017-10-20
Allow WebIndexPlugin to take template data also


pyros2097:
I reckon most of my projects to use the same index.html code. So would it be possible to provide the template inline instead of passing a file path to it. Will make it easier related to configuration since now I can also use string templates to inject keys for production and also this will remove the need for a single index.html file in my folder structure. Just an idea.
```html
<!DOCTYPE html>
<html>

<head>
  <title></title>
</head>

<style>
  body {
    margin: 0;
    padding: 0;
    font-family: sans-serif;
  }
</style>

<body>
  <div id="root"></div>
  $bundles
</body>

</html>
```

nchanged:
thanks @Shepless for implementing it so quick! You rock 🔥

PR #886, opened 2017-10-20
feat: WebIndexPlugin now accepts a template as a string literal

Shepless:
As requested by @pyros2097 in #885 . Its a simple enough change so I did it real quick. Allows you to do the following:

```js
plugins: [
   WebIndexPlugin({
      templateString: `Your stuff here`
   })
]
```

***NOTE:*** If you specify a `templateString` and `template` then `template` will take precedent


pyros2097:
Nice. So fast that too while polybar was still compiling on my system. On first I thought you were some old man with beard but on closer look that looks a lot like soap.

PR #890, opened 2017-10-21
feat: Allow extension overrides configuration

Shepless:
FuseBox can now be configured to look for customised extensions ahead of standard ones. This is
useful for platform specific builds. For example:

```js
FuseBox.init({
   extensionOverrides: ['.desktop.js']
});
```

With the above configuration FuseBox will favour loading contents from files with a `.desktop.js` extension over `.js` extension. If `.desktop.js` does not exist then normal resolution will occur. This a essentially a port of the Webpack functionality for [resolve extensions](https://webpack.js.org/configuration/resolve/#resolve-extensions).

In Summary:
- Added Functionality
- Added Unit Tests
- Added Documentation

### DIFF
--- a/src/plugins/stylesheet/SassPlugin.ts
+++ b/src/plugins/stylesheet/SassPlugin.ts
@@ -80,12 +80,20 @@ export class SassPluginClass implements Plugin {
                 if (/https?:/.test(url)) {
                     return done({ url });
                 }
+
                 for (let key in options.macros) {
                     if (options.macros.hasOwnProperty(key)) {
                         url = url.replace(key, options.macros[key]);
                     }
                 }
-                done({ file: path.normalize(url) });
+
+                let file = path.normalize(url);
+
+                if (context.extensionOverrides) {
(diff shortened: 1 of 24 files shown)


src/plugins/stylesheet/SassPlugin.ts
```diff
@@ -80,12 +80,20 @@ export class SassPluginClass implements Plugin {
                 if (/https?:/.test(url)) {
                     return done({ url });
                 }
+
                 for (let key in options.macros) {
                     if (options.macros.hasOwnProperty(key)) {
                         url = url.replace(key, options.macros[key]);
                     }
                 }
-                done({ file: path.normalize(url) });
```
nchanged: Why did you remove the normalization?

Shepless: It's not been removed - it does it still, see line 90. It normalises then it checks for overrides 👍

nchanged:
Thanks a billion @Shepless 
Really good job!

Issue #860, opened 2017-10-11
[PostCSSPlugin] Unexpected token o in JSON at position 1


cmizzi:
Hi everyone.

I'm currently testing `fuse-box` with `PostCSSPlugin` and `CSSPlugin`. My goal is the following : 

* creates an `built/app.css` with my custom grouped CSS files (apply PostCSS on this file and prevents inject) ;
* injects the rest of the CSS that can be loaded dynamically but is provided by vendor packages (just CSS) ;

This is working, but when I want `sourceMaps: true`, I've got this kind of error

```bash
10:15:10: Source changed for index.js
10:15:10: Source changed for index.js
SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
    at File.generateCorrectSourceMap (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/core/File.js:438:39)
    at group.subFiles.forEach.file (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/plugins/stylesheet/CSSplugin.js:51:67)
    at Array.forEach (native)
    at CSSPluginClass.transformGroup (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/plugins/stylesheet/CSSplugin.js:49:24)
    at promises.push.Promise (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/core/ModuleCollection.js:144:59)
    at context.fileGroups.forEach (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/core/ModuleCollection.js:143:35)
    at Map.forEach (native)
    at ModuleCollection.transformGroups (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/core/ModuleCollection.js:139:33)
    at realm_utils_1.each.then.then.then.then.then (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/core/ModuleCollection.js:92:34)

--------------------------
Bundle "app" 

    index.js
    style/app.css
    app.css
└──  (3 files,  543 Bytes) default
size: 543 Bytes in 30ms
SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
    at File.generateCorrectSourceMap (/data/cmizzi/fusebox/node_modules/fuse-box/dist/commonjs/core/File.js:438:39)
[...]

PR #897, opened 2017-10-24
fix: css plugin doesnt check type before calling generateCorrectSourc…

mwmessin:
generally it is bad to reuse variables with differently typed values. This bug is an example of that: 

```
    public generateCorrectSourceMap(fname?: string) {
        if (this.sourceMap) {
            let jsonSourceMaps = JSON.parse(this.sourceMap);
```

where the if isn't going to stop when this.sourceMap is assigned to an object. Alternatively you may want to do this typecheck in File.ts.

### DIFF
--- a/src/plugins/stylesheet/CSSplugin.ts
+++ b/src/plugins/stylesheet/CSSplugin.ts
@@ -211,7 +211,7 @@ export class CSSPluginClass implements Plugin {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


mwmessin:
This fixes issue https://github.com/fuse-box/fuse-box/issues/860

nchanged:
Thanks, as a temp solution that will work

Issue #916, opened 2017-11-02
es6 module in node_modules not transpiled


guillaume86:
I have a dependency of a dependency (`only-react-css-property-operations` which is required by `react-dom-animated`) which use the `const` es6 feature and it is not getting transpiled to es5 `var` when using the browser target (I also tried browser@es5).

I managed to force the transpilation by using a plugin to set `file.es6module = true` on the appropriate file.

From what I gathered from the source code, only files using es6 imports/exports are currently transpiled:
https://github.com/fuse-box/fuse-box/blob/master/src/analysis/plugins/ImportDeclaration.ts#L28-L46

Edit: to clarify, the module is using es6 syntax features (const), but not the es6 module (import/export) syntax.

nchanged:
@guillaume86 if you are familiar with AST you could try adding const let and other conditions here
https://github.com/fuse-box/fuse-box/blob/master/src/analysis/plugins/LanguageLevel.ts#L4

That's a very simple fix, i don't remember exactly the node name, but it looks something llke this:
```
if( node.kind === "const"){
     file.setLanguageLevel(ScriptTarget.ES2016)
}
```

Do you think you can make a PR?

guillaume86:
Ok I'm going to give it a shot.

Shepless:
@guillaume86 @nchanged can this be closed now the PR has been merged?

guillaume86:
Yes it's solved.

Le dim. 5 nov. 2017 17:31, Chris Shepherd <notifications@github.com> a
écrit :

PR #917, opened 2017-11-02
fix: LanguageLevel doesn't detect es2015 "const", "let" and "arrow function"

guillaume86:
As discussed in issue #916, I added detection of `const` and arrow functions `(...) => ...` since they are probably the most commonly used ES2015 features.

For complete language support, it would probably be better to create and use a separated module dedicated to this (something like 'acorn-detect-ecmaversion'), I didn't find anything existing covering this.

### DIFF
--- a/src/analysis/plugins/LanguageLevel.ts
+++ b/src/analysis/plugins/LanguageLevel.ts
@@ -4,6 +4,10 @@ export class LanguageLevel {
     public static onNode(file: File, node: any, parent: any) {
         if (node.async === true) {
             file.setLanguageLevel(ScriptTarget.ES2017)
+        } else if (node.kind === "const") {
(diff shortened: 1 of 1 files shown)


src/analysis/plugins/LanguageLevel.ts
```diff
@@ -4,6 +4,10 @@ export class LanguageLevel {
     public static onNode(file: File, node: any, parent: any) {
         if (node.async === true) {
             file.setLanguageLevel(ScriptTarget.ES2017)
+        } else if (node.kind === "const") {
```
nchanged: Shall we add `let` too?

guillaume86:
I added the "let" support and squashed the 2 commits together.

PR #969, opened 2017-12-04
fix: require function undefined access

andrew-w-ross:
cache isn't always defined so `cached = cache[this.filePath]` breaks the require function. This adds a check to fix it.

### DIFF
--- a/src/FuseProcess.ts
+++ b/src/FuseProcess.ts
@@ -31,8 +31,9 @@ export class FuseProcess {
 				mdl;
 		}
 		return new Promise((resolve, reject) => {
-			let cache = (<any>require).cache, cached = cache[this.filePath], closePromise, exps;
-			if (cached) {
+			let cache = (<any>require).cache, closePromise, exps
+			const cached = cache && cache[this.filePath];
+			if (cached != null) {
 				try {
 					if (opts.close)	//if a close function is given in parameter
 						closePromise = opts.close(cached.exports);


src/FuseProcess.ts
```diff
@@ -31,8 +31,9 @@ export class FuseProcess {
 				mdl;
 		}
 		return new Promise((resolve, reject) => {
-			let cache = (<any>require).cache, cached = cache[this.filePath], closePromise, exps;
-			if (cached) {
+			let cache = (<any>require).cache, closePromise, exps
+			const cached = cache && cache[this.filePath];
+			if (cached != null) {
```
nchanged: Null is quite a tricky thing, are u sure about it?

andrew-w-ross: Well I was just checking for null or undefined hence the `!=` and not `!==`. Unless you are referring to something else.

nchanged: What are the possible values of the `cache` variable?

andrew-w-ross: I have no idea cache is defined as any.

andrew-w-ross: @nchanged can you merge or reject and close this pr?

nchanged: Please don’t rush me on this one, I need to test it myself I just haven’t had time yet

nchanged:
```
var cache = undefined;
cache == null // true o_O
```

what's wrong with just `if (!cache) {}` it shouldn't yield falsy values

andrew-w-ross:
Seems to have been fixed in later releases

Issue #948, opened 2017-11-20
Imported scss file by other scss when changed, Fuse-box not bundling changes


serkon:
i have **main.ts** and it is import **main.scss**. Main.scss imports my **imports.scss**.

main.ts ####
import '~/main.scss';

main.scss ####
@import 'src/assets/styles/imports';

imports.scss ####
@import "config";  // --> when change something inside this file
@import "mixins";
@import "reset";
@import "font";
@import "grid";
........

when i change something inside config.scss, this change not bundling to project. If i change main.scss it was bundling.

node v9.2.0
"fuse-box": "^3.0.0-next.15"
"node-sass": "^4.7.1"

Shepless:
@serkon do you have a repo available so we can investigate a bit more?

serkon:
@Shepless hi,
I create angular.io project with fuse-box. When you run npm start the project it will redirect to sample page. Then please inspect 'sample text' element. You will see the .text{...} style class in Elements/Styles tab of developer toolbar. Please click to <style>…</style> right top of corner. It will redirect to inline css. 

But it must be redirect you to find-me.scss. I wish i explained it.

serkon:
Another situation, if you change something inside find-me.scss i expect to see on browser when hmr() is completed. But nothing has been changing. It will be change when restart to "npm start" or change something in main.scss

nchanged:
### FuseBox v4 Major Upgrade

As of today, the latest major version of FuseBox `v4` is available in `fuse-box@next`. FuseBox has undergone a full re-write from scratch, learning from its mistakes in the previous versions.

There are a lot of improvements, speed and config wise.

Get started [here](https://github.com/fuse-box/fuse-box/blob/master/docs/getting-started/get-started.md)

We are still working on the documentation and polishing before the major release. We are getting very close. And we need help testing.

```bash
npm install fuse-box@next
```

Most of the issues that I am closing have been resolved, however, if you find that your issue is still relevant in `v4` don't hesitate to re-open or (better) create a new one with `v4` configuration.

PR #974, opened 2017-12-06
fix: issue #948 css bundling

phamtm:
This is to fix https://github.com/fuse-box/fuse-box/issues/948. The root cause is that if you have a nested css structure like so:

a.scss
```scss
@import "./b.scss"
```

b.scss
```scss
// some css
```
Then somehow `b` is not recorded as a subfile of `a`, but is is available in `File.cssDependencies`. I'm not sure whether this is the right way `!!subFile.cssDependencies.find(dep => dep.indexOf(lastChangedFile) !== -1)`. Any pointer would be appreciated

### DIFF
--- a/src/plugins/stylesheet/CSSplugin.ts
+++ b/src/plugins/stylesheet/CSSplugin.ts
@@ -137,7 +137,16 @@ export class CSSPluginClass implements Plugin {
                   emitRequired = true;
               }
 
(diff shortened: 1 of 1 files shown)


nchanged:
@troynt could it be related to the fact that you don't need to set extensions? 

```
@import "./b"
```

PR #999, opened 2018-01-02
Support HMR of files included through CSS Modules @value

mdlawson:
The CSS Modules spec has a syntax for defining values which can be imported from within a css or js file. This syntax is supported by default via the postcss-modules library, so it works well for the first compile with fusebox. However, dependencies introduced by `@value` imports within CSS aren't tracked, so changes to the values themselves are not applied until the importing file is reloaded from some other change.

For example, 

main.css:
```css
@value primary from "./colors.css";

body {
  background: primary
}
```
colors.css:
```
@value primary: red;
```

If colors.css is modified so that primary is green, main.css is not reloaded and the change is not applied. This PR fixes that behavior. I also noticed that the PostCSSPlugin mutates its `paths` param, which is fixed.

I'm not 100% sure on this best design for this change, so this PR mainly follows patterns in other CSS plugins. I'd like for example to avoid passing the same `paths` parameter to both `CSSModules` and `PostCSSPlugin`, and it seems like there's now a lot of duplication between the two plugins in general. I'm also not sure about adding CSS Modules specific syntax to the `CSSDepedencyExtractor`.

Is this something you'd want to support, and if so is this a reasonable implementation?

### DIFF
--- a/src/plugins/stylesheet/CSSModules.ts
+++ b/src/plugins/stylesheet/CSSModules.ts
@@ -6,6 +6,7 @@ import * as postcss from "postcss";
 export interface CSSModulesOptions {
     useDefault?: boolean;
     scopedName?: string;
+    paths?: string[],
     root?: string;
 }
 
@@ -32,8 +33,25 @@ export class CSSModulesClass implements Plugin {
 
     public transform(file: File): Promise<any> {
         file.addStringDependency("fuse-box-css");
+
+        if (file.isCSSCached("cssmodules")) {
+            return;
+        }
+        file.bustCSSCache = true;
+
         return new Promise((resolve, reject) => {
             file.loadContents();
+            const context = file.context;
+
+            const paths = [file.info.absDir, ...this.options.paths || []]
+
+            const cssDependencies = context.extractCSSDependencies(file, {
+                paths: paths,
+                content: file.contents,
+                extensions: ["css"]
+            });
+            file.cssDependencies = cssDependencies;
+
             return postcss([
                 require('postcss-modules')({
                     root: this.options.root || file.info.absDir,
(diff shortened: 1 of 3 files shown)


src/plugins/stylesheet/CSSModules.ts
```diff
@@ -32,8 +33,25 @@ export class CSSModulesClass implements Plugin {
...
+        }
+        file.bustCSSCache = true;
+
         return new Promise((resolve, reject) => {
             file.loadContents();
+            const context = file.context;
+
+            const paths = [file.info.absDir, ...this.options.paths || []]
+
+            const cssDependencies = context.extractCSSDependencies(file, {
```
nchanged: Don't you think this has to be optional? Some people combine it with SassModule where it has its own dependency tracker.

mdlawson: I figured this situation would happen already when using sass but then using postcss for autoprefixing. If both dependency trackers can't be active at the same time, can this be handled internally somehow? I don't think more config is helpful.

nchanged: Can you verify that on your branch?

nchanged: @mdlawson the release for 3.0 is coming tomorrow, let's verify it to avoid bugs. And thank you very much!

mdlawson: I'll try check this evening. If this can't make the 3.0 release though that's fine with me.

nchanged: I will try verifying it today too, if I find time, I would like that to be in 3.0 ;-)

mdlawson: I've verified the combination of this change and the sass plugin with [this setup](mdlawson/fuse-box-css-modules-example). Changes to both the sass only `_partial.scss` and the css module value included from `colors.scss` both correctly result in hot reloads and expected console output. Is there anything else you'd like me to check out?

I did notice the text logged to the console even without this change is like:
```
  → CSS Dependency: _partial.scss depends on main.scss
```

Which seems the wrong way round to me. `main.scss` depends on `_partial.scss` which is why a change to `_partial.scss` needs to trigger a reload of `main.scss`

nchanged: hm, but you do have main.css -> _partial.css So main.css is a parent file with the dependency `_partial.css`

nchanged: Anyway thank you very much for this PR!

Issue #990, opened 2017-12-22
ScriptTarget ESx/ESyyyy aliases are wrong


jollytoad:
The aliases from ES(edition) to ES(year) are mismatched

Should be:
ES6 = ES2015  (ie. 6th editor is ES2015)
ES7 = ES2016
ES8 = ES2017

This enum is wrong:

```
export enum ScriptTarget {
    ES5 = 1,
    ES2015 = 2,
    ES2016 = 3,
    ES6 = 3,
    ES2017 = 4,
    ES7 = 4,
    ESNext = 5
}
```

should be:

```
export enum ScriptTarget {
    ES5 = 1,
    ES2015 = 2,
    ES6 = 2,
    ES2016 = 3,
    ES7 = 3,
    ES2017 = 4,
    ES8 = 4,
    ESNext = 5
}
```

jollytoad:
Probably best to not have ES8 (and discourage use of ES6/7 in favour of year)

https://www.typescriptlang.org/docs/handbook/compiler-options.html lists...

ES5 
ES6 
ES2015 
ES7 
ES2016 
ES2017 
ESNext

nchanged:
Good catch! @jollytoad could you submit a PR? (you can include es8) And the docs too, I would appreciate it greatly.
Thanks

PR #1001, opened 2018-01-03
fix: ScriptTarget ES-edition to ES-year aliases (#990)

jollytoad:
I've also set the default language level to ES2016 for compatibility with the old ES6 default which was actually ES7/ES2016. If the default should actually be ES6 then it may be best to use ES2015 form - avoiding the ESx aliases entirely within the codebase.

(It's so easy to get the ES year mixed up - i'm doing it myself now, you may notice from the extra commit!!!)

Added some notes to docs regarding aliases and recommendation to use year form, and state the default.

Personally i'd ditch the es-edition aliases entirely and go with the es-year form always, but i don't know if you need them for possible TypeScript configuration compatibility.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -541,7 +541,7 @@ Possible levels
 
 Aliases `es6` & `es7` are provided to align with the configuration options of TypeScript, but we recommend using the es-year form instead.
 
-The default language level is `es2017`
+The default language level is `es2016`
 
 
 ### Example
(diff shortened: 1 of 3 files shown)


nchanged:
Thank you so much! I will hold off merging for now, as I am preparing the release of 3.0, and there might be conflicts. `configuration.md` is moved here:
https://github.com/fuse-box/fuse-box/blob/doc-ver-3/docs/development/configuration.md

We need to fix that, The release is coming (hopefully) tomorrow

PR #1012, opened 2018-01-07
Fix for uglify es6 option being ignored in quantum

### LINKED ISSUE #1010: uglify: {es6: true} on QuantumConfig is ignored
When creating a TS project with target browser@es5 and then using uglify: {es6: true} inside the QuantumConfig object then fusebox is still using uglify-js instead of uglify-es, which makes it fail since a dependency uses es6.

xaviergonz:
fixes #1010


xaviergonz:
No problem :)

nchanged:
Just for the future @xaviergonz please do npm install, we use commitizen, so all commits should follow the standard like `fix:` `feat:`  that's how we generate changelog

PR #1017, opened 2018-01-08
Allow relative URLs in less files

ccapndave:
This PR adds the `relativeUrls` option to `LESSPlugin`.  Its necessary when you want to automatically rebase urls in imported less files relative to the entry file.  The previous fix for filename is actually only ever used (AFAIK) when `relativeUrls` is turned on so I believe nothing should be broken by that.

### DIFF
--- a/src/plugins/stylesheet/LESSPlugin.ts
+++ b/src/plugins/stylesheet/LESSPlugin.ts
@@ -1,6 +1,7 @@
 import { File } from "../../core/File";
 import { WorkFlowContext } from "./../../core/WorkflowContext";
 import { Plugin } from "../../core/WorkflowContext";
+import * as path from "path";
 let less;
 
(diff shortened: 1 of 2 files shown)


src/plugins/stylesheet/LESSPlugin.ts
```diff
@@ -56,7 +57,7 @@ export class LESSPluginClass implements Plugin {
             less = require("less");
         }
 
-        options.filename = file.context.homeDir + (options.filename || file.info.fuseBoxPath);
+        options.filename = options.filename ? file.context.homeDir + options.filename : file.info.absPath;
```
nchanged: You think we can `path.join` it here?

ccapndave: For sure

PR #1037, opened 2018-01-15
fix: Corrects updates for HMR updates by pointing to the correct map file

### LINKED ISSUE #1029: HMR update breaks source maps in vscode
On update when the a `source-changed` event fires the sourceMapping url is incorrect as it points to `//# sourceMappingURL=${nameOfFileThatWasChange}.map` which doesn't exist and therefor breaks when vscode tries to map back to the original map file. 

You can recreate it using the [fuse-debug](/andrew-w-ross/fuse-debug#master) that uses v2 there is also a v3 branch that shows the same behaviour. 

I think a fix would be quite simple as the sourceMappingURL should just be the bundle name. I know there is a lot on your plate after the release of v3 so I'm more than happy to make a pr.

andrew-w-ross:
Adds a guide on how to debug in vscode
Corrects updates for hmr updates by pointing to the correct map file closes #1029


docs/guides/debugging-with-vscode.md
```diff
@@ -0,0 +1,88 @@
+# Debugging with vscode
+
+[Visual Studo Code](https://code.visualstudio.com/) is a light weight editor, with excellent debugging support. In this guide you'll be shown on how to get debugging working with `fuse-box` on both web apps and node applications. The source code for this example application can be found [here](/andrew-ross/fuse-debug).
```
nchanged: I think a full link is required, as we are hosting it
`/andrew-ross/fuse-debug`

docs/guides/debugging-with-vscode.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+## Browser Debugging
+
+Make sure the [debugger for chrome](https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome) is installed. 
+Then Start up fuse box by running `npm start` in the terminal and wait for the `Listening @ http://localhost:3000` message. In vscode open the debug panel and select the `Launch Chrome` configuration from the drop down and launch by pressing `f5`.
+
+Add a break point to either the `render` or `add` function in the `Adder.tsx` and hit the Click Me button. It should break as demoed below:
+
+![debugging gif](../images/debugging.gif)
```
nchanged: also here, full link otherwise it won't be visible on the website

docs/guides/debugging-with-vscode.md
andrew-w-ross: Done

docs/guides/debugging-with-vscode.md
andrew-w-ross: Done

src/core/File.ts
```diff
@@ -645,9 +645,9 @@ export class File {
 
             if (!this.context.inlineSourceMaps) {
                 delete jsonSourceMaps.sourcesContent;
-            }
-
-            result.outputText = result.outputText.replace("//# sourceMappingURL=module.js.map", "");
```
nchanged: I still think we need to keep the original replacement. As TypeScript adds it to each file. Can we keep it too?

PR #1079, opened 2018-01-29
feat: bake quantum api into multiple bundles

CzBuCHi:
didnt create any tests, because im not sure where to put them (there are none for original bakeApiIntoBundle behavior)


CzBuCHi:
replaced by #1086

PR #1107, opened 2018-02-15
Fix error on VueComponentPlugin and optimize plugins on electron

orblazer:
Hello, i have fix error on VueComponentPlugin and optimize plugins on electron.

Changes :
-  Fix error "Uncaught Package not found process" when use VueComponentPlugin in development.
This error is produced because VueComponentPlugin call `FuseBox.import("process")` before define `FuseBox.target`.

    This fix, fix probably other bugs when we need falls functions before `FuseBox.target`, `FuseBox.isServer` and `FuseBox.sdep`.
    For fix that i have just move this lines to start bundle.
- Add verry small memory save on FuseBox loader when check if is browser (variable : `$isBrowser`). This variable now store only boolean and not reference to `window.navigator`
- Optimize VueComponentPlugin when running on electron or server.
For this, i have replace `FuseBox.import('process')` by `global.process`.
And i have replace `process.env.vueHMR` by `__VUE_HOT_MAP__` (this is temporary because i wait this PR is accepted https://github.com/vuejs/vue-hot-reload-api/pull/66)
This is fix a bug when is runned on electron because the vueHMR is string serialized.
- Optimize EnvPlugin when running on electron.
For this i have replace variable `__process_env__` by directly `global.process.env`, like server.

Thanks for your work !

### DIFF
--- a/src/plugins/vue/VuePlugin.ts
+++ b/src/plugins/vue/VuePlugin.ts
@@ -94,10 +94,11 @@ export class VueComponentClass implements Plugin {
 
   public bundleEnd(context: WorkFlowContext) {
     if (context.useCache && this.hasProcessedVueFile) {
-      context.source.addContent(`
-        var process = FuseBox.import('process');
-
-        if (process.env.NODE_ENV !== "production") {
+      const haveGlobal = context.isGlobalyIgnored('process')
+      const process = haveGlobal ? 'global.process' : 'process'
+      const importProcess = haveGlobal ? '' : 'var process = FuseBox.import(\'process\');\n\n'
+      context.source.addContent(`${importProcess}
+        if (${process}.env.NODE_ENV !== "production") {
           var api = FuseBox.import('vue-hot-reload-api');
           var Vue = FuseBox.import('vue');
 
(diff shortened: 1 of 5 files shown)


nchanged:
@orblazer great job!!
Can I ask you to amend the commits, so they will actually follow the commit standards? 
as in:

```
fix: Fixes bugs when using VueComponentPlugin in development.
fix: Optimize component HMR
```

These commits go to our changelog, so they will have to be as accurate as possible, otherwise, I will be forced to squash it ;-(

src/plugins/vue/VuePlugin.ts
```diff
@@ -94,10 +94,11 @@ export class VueComponentClass implements Plugin {
 
   public bundleEnd(context: WorkFlowContext) {
     if (context.useCache && this.hasProcessedVueFile) {
-      context.source.addContent(`
-        var process = FuseBox.import('process');
-
-        if (process.env.NODE_ENV !== "production") {
+      const haveGlobal = context.isGlobalyIgnored('process')
```
nchanged: nitpick: `hasGlobal`?

src/core/ServerPolyfillList.ts
```diff
@@ -49,9 +49,9 @@ const ELECTRON_POLYFILL = new Set<string>(
     ]
 );
 export function isServerPolyfill(name: string) {
-    return SERVER_POLYFILL.has(name);
+    return SERVER_POLYFILL.has(name) || false;
```
nchanged: I think `Set.has`returns a boolean by default

orblazer: Hum yep i am stupid i go revert the commit and replace by real fix

src/plugins/vue/VuePlugin.ts
orblazer: hep i don't know i have write "have"

orblazer:
Ok, now the PR is probably valid.

nchanged:
I think the last commit can be squashed into `fix: Optimize VuePlugin when is running on electron`

orblazer:
Ok, no problem i fix than in this hour

orblazer:
Ok, is done !

src/plugins/vue/VuePlugin.ts
```diff
@@ -94,9 +94,9 @@ export class VueComponentClass implements Plugin {
 
   public bundleEnd(context: WorkFlowContext) {
     if (context.useCache && this.hasProcessedVueFile) {
-      const haveGlobal = context.isGlobalyIgnored('process')
-      const process = haveGlobal ? 'global.process' : 'process'
-      const importProcess = haveGlobal ? '' : 'var process = FuseBox.import(\'process\');\n\n'
+      const hasGlobal = context.isGlobalyIgnored('process')
+      const process = hasGlobal ? 'global.process' : 'process'
+      const importProcess = hasGlobal ? '' : 'var process = FuseBox.import(\'process\');\n\n'
```
nchanged: Just noticed why do we need to define process globally?

orblazer: I use global process from global because the simple variable process is not defined, this is strange because when i test with `this` context that is work.. I don't know the reason of the bug.

An probably solution it is add call the anonymous function with process argument.

PR #1111, opened 2018-02-16
fix: Added react and react-dom to devDependencies

deamme:
I'm having some conflicting issues when developing https://github.com/deamme/laco. I don't think fuse-box is using `react` and `react-dom` internally so this should be fine.


deamme:
Just moved `react` and `react-dom` to `devDependencies ` instead of `dependencies` in `package.json`

nchanged:
Ooops, they snuck in there by accident

PR #1120, opened 2018-02-21
Update the condition for vue HMR

orblazer:
Hello, i have add function `isRecorded` on vue-hot-reload-api ([this PR](https://github.com/vuejs/vue-hot-reload-api/pull/66)).

Now the VueComponentPlugin require `vue-hot-reload-api@^2.3.0`.

I have two question, would not it be better to add plugin dependencies in `peerDependencies`.
Do you want replace `ws` lib by `uws` for have more performance on HMR and less dependencies. 
End for end, the `watch` dependencies is not used ?
Yep i search how to reduce dependencies size :)

Thanks.

### DIFF
--- a/src/plugins/vue/VuePlugin.ts
+++ b/src/plugins/vue/VuePlugin.ts
@@ -269,7 +269,7 @@ export class VueComponentClass implements Plugin {
         if (${process}.env.NODE_ENV !== "production") {
           var api = require('vue-hot-reload-api');
 
-          if (__VUE_HOT_MAP__ && !__VUE_HOT_MAP__['${moduleId}']) {
+          if (api && !api.isRecorded('${moduleId}')) {
             api.createRecord('${moduleId}', module.exports.default);
           }
         }


nchanged:
Great! There are quite many peerDeps, I am not sure if adding all of them is a good idea.
`uws` that's a library? We can, but there are many places that would require refactoring.

nchanged:
Great! There are quite many peerDeps, I am not sure if adding all of them is a good idea.
`uws` that's a library? We can, but there are many places that would require refactoring.

orblazer:
Ok, if you want i can make the refactoring in next days.

nchanged:
@orblazer You have to be very careful with that one. You can't break the API

orblazer:
Yes, i know.

nchanged:
@orblazer Try to make it simpler to review, I will be very strict with it because this change can affect a lot of projects. ;-)

nchanged:
And BTW, thanks a lot for contributing! I really appreciate it! I hope we can get more people onboard!

PR #1124, opened 2018-02-22
refactor: Change WS library by UWS

orblazer:
Hello,
I have replace `ws` library by `uws` for increase performance.

The `uws` library is very similar to `ws` and he don't need refactoring codes.

Thanks.

PR #1125, opened 2018-02-22
Clean dependencies and fix tests

orblazer:
Hello,
I have clean unused dependencies and depcreated.

The "refactor: Change WS library by UWS" wrong here, the PR #1124 need merged before.

Thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fuse-box",
-  "version": "3.2.0-next.4",
+  "version": "3.2.0-next.6",
   "description": "Fuse-Box a bundler that does it right",
   "typings": "index.d.ts",
   "main": "index.js",
@@ -32,6 +32,7 @@
   "devDependencies": {
     "@commitlint/cli": "^6.1.0",
     "@commitlint/config-conventional": "^6.1.0",
+    "@types/comment-json": "^1.1.1",
     "@types/node": "^6.0.74",
     "cheerio": "^0.22.0",
     "commitizen": "^2.9.6",
@@ -71,6 +72,7 @@
     "base64-js": "^1.2.0",
     "chokidar": "^1.6.1",
     "clean-css": "^4.1.9",
+    "comment-json": "^1.1.3",
     "escodegen": "^1.8.1",
     "express": "^4.14.0",
     "fliplog": "^0.3.13",
@@ -86,6 +88,7 @@
     "realm-utils": "^1.0.9",
     "request": "^2.79.0",
     "shorthash": "0.0.2",
+    "source-map": "^0.7.1",
     "tslib": "^1.8.0",
     "uws": "^9.14.0",
     "yargs": "^9.0.1"
--- a/docs/development/configuration.md
+++ b/docs/development/configuration.md
@@ -68,6 +68,12 @@ Will be converted accordingly to:
 
 However it is strongly recommended to use `$name` in your output template, as `out.js` will be ignored
 
+## stdin
+
+Listens to stdin.
+
+Combinations like `CMD+D` will gracefully shut down any process.
+
 ## Cache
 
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -82,19 +78,15 @@
...
     "ieee754": "^1.1.8",
-    "inquirer": "^3.0.6",
     "lego-api": "^1.0.7",
     "mustache": "^2.3.0",
     "postcss": "^6.0.1",
     "pretty-time": "^0.2.0",
     "prettysize": "0.0.3",
     "realm-utils": "^1.0.9",
-    "regexpu-core": "^4.1.3",
-    "request": "^2.79.0",
```
nchanged: We need request, Quantum uses it with one option

src/devServer/SocketServer.ts
```diff
@@ -1,5 +1,5 @@
 import { FuseBox } from "../core/FuseBox";
-import { Server } from "ws";
```
nchanged: I think that belongs to the different PR ;-)

package.json
```diff
@@ -82,19 +78,15 @@
     "fuse-concat-with-sourcemaps": "^1.0.5",
     "glob": "^7.1.1",
     "ieee754": "^1.1.8",
-    "inquirer": "^3.0.6",
```
nchanged: Are you sure it's not in use anywhere?

src/devServer/SocketServer.ts
orblazer: yep but i dont know how to skip this commit :(

package.json
orblazer: Yep, i have search in all files on github and in the project

package.json
orblazer: Hump i don't find the usage in another place than test

nchanged: https://github.com/fuse-box/fuse-box/blob/8aa160b142e6a3d069b192aa19b271e47e329d2a/modules/fuse-box-responsive-api/index.js#L187-L194

src/devServer/SocketServer.ts
nchanged: I will check your first PR soon, then you can rebase

package.json
nchanged: 👍

package.json
orblazer: ops, ok i rewrite commit history for fix that

orblazer:
Ok, i have rewrite the history of commits for cleanup.
Normally this PR is OK.

orblazer:
Hum... i have an problem, the test "Should replace an alias - as a file - with babel" not pass because the alias is not apply.

For begin fix i have replace that condition [ImportDeclaration.ts#L40](https://github.com/fuse-box/fuse-box/blob/master/src/analysis/plugins/ImportDeclaration.ts#L40) by :
```ts
if (node.source && analysis.nodeIsString(node.source)) {
    let requireStatement = this.handleAliasReplacement(file, node.source.value);
    if (requireStatement) {
        analysis.registerReplacement(node.source.value, requireStatement);
        node.source.value = requireStatement;
        analysis.addDependency(requireStatement);
    }
}
```
But now, this line in [FileAnalysis.ts#L150](https://github.com/fuse-box/fuse-box/blob/master/src/analysis/FileAnalysis.ts#L150) not apply the alias because the content of file is :
```js
export { default as canada } from 'eh';
```
We probably need change the regex but i don't know where to replace her.

orblazer:
Hum i don't know, probably because i have update `babel-preset-latest` to `babel-preset-env` and the output code is ES2015. Probably before that the output code is ES5 and not ES6.

nchanged:
@orblazer can you cleanup this PR?

orblazer:
Yes, but i close this PR and make new clean PR.

PR #1129, opened 2018-02-24
fix: Support tsconfig.json with comments

cjdell:
I was having issues using the default `tsconfig.json` provided by `tsc --init`. This file includes comments which are not support by NodeJS's default JSON parser. I've switched to an alternative parser to support JSON comments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "wires-reactive": "^1.1.1"
   },
   "dependencies": {
+    "@types/comment-json": "^1.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "wires-reactive": "^1.1.1"
   },
   "dependencies": {
+    "@types/comment-json": "^1.1.1",
```
nchanged: Can you move this one to dev deps, please?

PR #1136, opened 2018-03-01
feat: log option now takes an object

ccapndave:
Adds some useful logging options.  Note that this is not backwards compatible, so anyone using `log: true` or `log: false` will need to change their config or will get a type error.

### DIFF
--- a/src/core/FuseBox.ts
+++ b/src/core/FuseBox.ts
@@ -37,7 +37,7 @@ export interface FuseBoxOptions {
      * default: "universal@es5"
      */
     target?: string,
-    log?: boolean;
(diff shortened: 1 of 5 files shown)


ccapndave:
I'm actually not 100% convinced that the bit that clears the terminal is in the right place.  Is there somewhere better for it?  Really I suppose I want it to clear when a watch fires.

ccapndave:
Actually even `showBundledFiles: false` doesn't seem to always work; occasionally it still lists the files.

src/core/FuseBox.ts
```diff
@@ -37,7 +37,7 @@ export interface FuseBoxOptions {
      * default: "universal@es5"
      */
     target?: string,
```
nchanged: Oh can you add | boolean to allow just booleans

PR #1144, opened 2018-03-05
Possibility to shape how bundles are emited

tomitrescak:
I have a scenario in which I am only loading JS if I am on a specific browser, supporting ES6 with async await. For all other browsers I am disabling the app. It is a corporate app so don't judge.

I can now do:

```html
<script type="text/javascript">
            function isGeneratorSupported() {
                try {
                    eval('(function*(){})()');
                    return true;
                } catch (err) {
                    return false;
                }
            }
            let c = document.getElementById('hiddenContent');
            if (!isGeneratorSupported() || /MSIE \d|Trident.*rv:/.test(navigator.userAgent)) {
                c.style.display = 'block';
            } else {
                $bundles
            } 
        </script>
```

and

```js
WebIndexPlugin({
        template: 'src/client/index.html',
        emitBundles: bundles =>
          bundles
            .map(b => `document.write('<script type="text/javascript" src="${b}"><\\/script>');`)
            .join('\n')
      }),
```

### DIFF
--- a/src/plugins/WebIndexPlugin.ts
+++ b/src/plugins/WebIndexPlugin.ts
@@ -21,6 +21,7 @@ export interface IndexPluginOptions {
     async?: boolean;
     pre?: { relType: 'fetch' | 'load' };
     resolve?: { (output: UserOutput): string };
+    emitBundles?: (bundles: string[]) => string;
 }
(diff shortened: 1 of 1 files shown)


nchanged:
@tomitrescak ?

PR #1177, opened 2018-03-29
feat: add warning on not found file

vemoo:
Fuse doesn't give any info when an imported css file for example is not found. This PR adds a warning for that case, I think that's the right place to put it


jpike88:
Just noticed my pr is related to this one.

@nchanged should I modify my one to become an 'else' statement? I don't think a warning is enough, my way of using this.addError makes it more obvious: you should never be building without your assets!

vemoo:
I agree it should be an error, but I used warning because that's what's being used for js files.

I we change it to error for js files it should also be error

jpike88:
@vemoo I've updated #1207 to basically do what your PR is doing... you can close this and follow the one if you want because that also tweaks another thing as well

jpike88:
@vemoo #1207 just got merged... can you close this pr to clean things up, cheers

PR #1207, opened 2018-04-18
Made error message clearer, and included HTML Template file not found error check.

### LINKED ISSUE #1204: Incorrect error message
This error assumes that TypeScript is not installed, and hides whatever error may be firing instead.

https://github.com/fuse-box/fuse-box/blob/e9f84099b93030ffcea9291e9f86f2388e2b99ca/src/core/File.ts#L622

screen shot 2018-04-16 at 9 16 49 pm

jpike88:
fixes #1204, and also includes an import check for importing HTML files with the ng-template plugin.

### DIFF
--- a/src/core/File.ts
+++ b/src/core/File.ts
@@ -451,6 +451,9 @@ export class File {
                 this.context.fuse.producer.addWarning('unresolved',
                     `Statement "${this.info.fuseBoxPath}" has failed to resolve in module "${this.collection && this.collection.name}"`);
             }
+            if (/\.html$/.test(this.info.fuseBoxPath)){
+                this.addError(`HTML Template Reference "${this.info.fuseBoxPath}" has failed to resolve in module "${this.collection && this.collection.name}"`);
+            }
             this.notFound = true;
             return;
         }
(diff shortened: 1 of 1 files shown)


jpike88:
@nchanged hopefully this helps!

jpike88:
Also, if you could make a release based on this that'd be sweet

src/core/File.ts
```diff
@@ -451,6 +451,9 @@ export class File {
                 this.context.fuse.producer.addWarning('unresolved',
                     `Statement "${this.info.fuseBoxPath}" has failed to resolve in module "${this.collection && this.collection.name}"`);
             }
+            if (/\.html$/.test(this.info.fuseBoxPath)){
```
nchanged: I don't think we need to need a special case for HTML. 
IT should be generic and message customised.

jpike88:
I also wonder if it's worth converting line 452 to an error instead of a warning as well. What do you think?

src/core/File.ts
```diff
@@ -450,6 +450,8 @@ export class File {
             if (/\.js$/.test(this.info.fuseBoxPath) && this.context.fuse && this.context.fuse.producer) {
```
nchanged: I think it can be `jsx` too. Or other things that should be interpreted as assets

jpike88: Line 454 catches all else statements, so .jsx will be caught. I wonder if all of them should be throwing an error?

nchanged: Yeah, just jsx isn't an assets, it's a module ;-)

jpike88: Done. Does it make sense for one to throw warnings and the else to throw errors? I would think that errors should always be thrown is something is missing?

PR #1223, opened 2018-04-26
Quantum fix

### LINKED ISSUE #1217: Quantum doesn't work: Globals only loading last variable in globals config
Using @next, found a weird bug in quantum that prevents my builds from working properly.

My globals config looks like this:

screen shot 2018-04-24 at 12 04 04 pm

I'm using bakeApiIntoBundle too just FYI.

The end of my app.js looks like this:

screen shot 2018-04-24 at 12 04 46 pm

If I change the order of the globals config, then the last one will be the only one that is exposed.

Will investigate further, but this is definitely a quantum related bug.

jpike88:
Allows a globals configuration object to function identically across quantum and non-quantum builds.

Fixes #1217

### DIFF
--- a/src/quantum/plugin/FlatFileGenerator.ts
+++ b/src/quantum/plugin/FlatFileGenerator.ts
@@ -5,8 +5,12 @@ import { BundleAbstraction } from "../core/BundleAbstraction";
 export class FlatFileGenerator {
     public contents = [];
     public entryId;
-    public globals: Map<string, number>;
+    public globals = new Map<string, string>();
     constructor(public core: QuantumCore, public bundleAbstraction?: BundleAbstraction​​) { }
+
+    public setGlobals(packageName: string, fileID: string) {
+        this.globals.set(packageName, fileID);
+    }
 
     public init() {
         if (this.core.opts.isTargetBrowser() || this.core.opts.isTargetUniveral()) {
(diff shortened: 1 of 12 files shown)


src/quantum/plugin/FlatFileGenerator.ts
```diff
@@ -68,25 +68,23 @@ export class FlatFileGenerator {
...
-            const req = `${this.core.opts.quantumVariableName}.r(${JSON.stringify(this.entryId)})`;
 
-            if (this.globalsName) {
-                if (this.core.opts.isTargetNpm() || this.core.opts.isTargetServer()) {
-                    this.contents.push(`module.exports = ${req}`);
-                }
+        this.globals.forEach((fileID, globalName) => {
+            const req = `${this.core.opts.quantumVariableName}.r(${JSON.stringify(fileID)})`;
+            if (this.core.opts.isTargetNpm() || this.core.opts.isTargetServer()) {
+                this.contents.push(`module.exports = ${req}`);
```
nchanged: It's not gonna work for module.exports here
We need to have only one module exports for server

nchanged: Also please check the tests

nchanged: They all need to pass, before we write new ones

jpike88: Fixed, also brought back in use of "*"

nchanged:
Awesome! I will check it manually today!

PR #1229, opened 2018-04-27
fix: require statements ignored if not 'true' requires. fixes #1142

### LINKED ISSUE #1142: Quantum not working well with variables named "require" inside AngularJS code
Greetings,
I'm having an issue very similar to this one: https://github.com/fuse-box/fuse-box/issues/1133
In AngularJS (1.6.9) there is this piece of code:

```
[...]
    function getControllers(directiveName, require, $element, elementControllers) {
      var value;

      if (isString(require)) {
        var match = require.match(REQUIRE_PREFIX_REGEXP);
        var name = require.substring(match[0].length);
[...]
```

Which doesn't get parse correctly when using quantum. 

Error: `ReferenceError: require is not defined`

Converted code (without uglify):  

```
function getControllers(directiveName, $fsx, $element, elementControllers) {
                    var value;
                    if (isString($fsx)) {
                        var match = require.match(REQUIRE_PREFIX_REGEXP);
                        var name = require.substring(match[0].length);

```

Reading the issue linked above i updated fusebox to 3.2.0-next10 but then something similar happened to another piece of code in AngularJS:

AngularJS code:

```
  function getDirectiveRequire(directive) {
    var require = directive.require || (directive.controller && directive.name);

    if (!isArray(require) && isObject(require)) {
      forEach(require, function(value, key) {
        var match = value.match(REQUIRE_PREFIX_REGEXP);
        var name = value.substring(match[0].length);
        if (!name) require[key] = match[0] + key;
      });
    }
```
Error: `TypeError: e.match is not a function`

Converted code (without uglify):  

```
function getDirectiveRequire(directive) {
            var require = directive.require || directive.controller && directive.name;
            if (!isArray($fsx) && isObject($fsx)) {
                forEach($fsx, function (value, key) {
                    var match = value.match(REQUIRE_PREFIX_REGEXP);
                    var name = value.substring(match[0].length);
                    if (!name)
                        require[key] = match[0] + key;
                });
            }
            return $fsx;
        }
```

This is my quantum configuration (i have tried with any combination of replaceTypeOf and mangle, also completely disabling uglify, still not working)
```

				this.isProduction && QuantumPlugin({
					//bakeApiIntoBundle: "app",
					replaceTypeOf: false,
					treeshake: true,
					uglify: {
						mangle: true
					},
					css: {
						clean: true
					},
					api: core => {
						core.solveComputed("moment/moment.js", {
							mapping: "moment/locale**",
							fn: statement => {
                                                             statement.setExpression(`"moment/locale/" + name + ".js"`);
							}
						});
					}
					// containedAPI : true
				})
```

jpike88:
fixes #1142

### DIFF
--- a/src/quantum/core/AstUtils.ts
+++ b/src/quantum/core/AstUtils.ts
@@ -267,7 +267,7 @@ function lookUpTreeForBodyWithRequireFlag(node, ignoreFlagCheck) {
 // if it still passes thorough, it's piped to the matchesSingleVariable function for good measure.
 export function isTrueRequireFunction(node) {
     if (node.type === "Identifier" && node.name === "require") {
-        const isVar = node.$parent && node.$parent.type === 'VariableDeclarator' && node.$parent.left === node;
+        const isVar = node.$parent && node.$parent.type === 'VariableDeclarator' && node.$parent.init !== node;
         const isParam = node.$prop == 'params';
         if (isVar || isParam) {
             // get nearest body up chain and set flag.
--- a/src/quantum/core/AstUtils.ts
+++ b/src/quantum/core/AstUtils.ts
@@ -49,7 +49,7 @@ export function matchesDefinedExpression(node, expressions: { [key: string]: boo
     let isConditional = false;
 
     if (node.type === "IfStatement" && node.test && node.test.type === "BinaryExpression"
-        && node.test.left && ( node.test.left.type === "MemberExpression" || node.test.left.type === "Identifier" ) ) {
+        && node.test.left && (node.test.left.type === "MemberExpression" || node.test.left.type === "Identifier")) {
         targetNode = node.test.left;
         isConditional = true;
     }
@@ -211,8 +211,8 @@ export function matchesSingleVariable(node: any, name: string) {
                 return false;
             }
             if (parent.type) {
-                if( parent.type === "UnaryExpression"){
-                    if ( parent.argument && parent.operator === "typeof" && parent.argument.type === "Identifier" && parent.argument.name === name){
+                if (parent.type === "UnaryExpression") {
+                    if (parent.argument && parent.operator === "typeof" && parent.argument.type === "Identifier" && parent.argument.name === name) {
                         return false;
                     }
                 }
@@ -232,8 +232,19 @@ export function matchesSingleVariable(node: any, name: string) {
 }
 
 
-export function matchesRequireFunction(node: any) {
-    return matchesSingleVariable(node, "require");
+export function isTrueRequireFunction(node: any) {
+    if (matchesSingleVariable(node, "require")) {
+        const isVar = ["var", "const", "let"].indexOf(node.type) > -1;
+        const isParam = node.$prop == 'params';
+        const hasParentWithFlag = node.$parent && node.$parent.skipRequireSubstitution;
+        if (isVar || isParam || hasParentWithFlag) {
+            node.$parent.skipRequireSubstitution = true;
+            node.skipRequireSubstitution = true;
+            return false;
+        } else {
+            return true;
+        }
+    }
 }
 
 export function matchesSingleFunction(node: any, name: string) {
--- a/src/quantum/core/AstUtils.ts
+++ b/src/quantum/core/AstUtils.ts
@@ -49,7 +49,7 @@ export function matchesDefinedExpression(node, expressions: { [key: string]: boo
     let isConditional = false;
 
     if (node.type === "IfStatement" && node.test && node.test.type === "BinaryExpression"
-        && node.test.left && ( node.test.left.type === "MemberExpression" || node.test.left.type === "Identifier" ) ) {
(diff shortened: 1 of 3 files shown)


src/quantum/core/AstUtils.ts
```diff
@@ -232,8 +232,19 @@ export function matchesSingleVariable(node: any, name: string) {
 }
 
 
-export function matchesRequireFunction(node: any) {
-    return matchesSingleVariable(node, "require");
+export function isTrueRequireFunction(node: any) {
+    if (matchesSingleVariable(node, "require")) {
```
nchanged: I think here should be something like:
```
if( node.$parent.skipRequireSubstitution ) {
  node.skipRequireSubstitution = true
}
```

in order to keep it nested

jpike88:
it's just been given a solid makeover, have another look

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -74,6 +74,60 @@ export class GlobalVariableTEst {
...
+                    `
+                    },
+                    plugins: [QuantumPlugin({
+                        target: "browser"
+                    })]
+                }
+            }
+        ).simple().then(test => test.browser((window, env) => {
+            const app = env.getScript("app.js");
+            app.shouldFindString("require");
```
nchanged: `isString(require)` ?

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -74,6 +74,60 @@ export class GlobalVariableTEst {
...
```
nchanged: app.shouldFindString("function(require)");
app.shouldFindString("isString(require)");
app.shouldNotFindString("$fsx");

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -140,4 +194,29 @@ export class GlobalVariableTEst {
...
+                    `
+                    },
+                    plugins: [QuantumPlugin({
+                        target: "browser"
+                    })]
+                }
+            }
+        ).simple().then(test => test.browser((window, env) => {
+            const app = env.getScript("app.js");
+            app.shouldNotFindString("require('test')");
```
nchanged: You are executing an invalid js here ;-) require is `2` should be a function instead

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -140,4 +194,29 @@ export class GlobalVariableTEst {
...
+        }));
+    }
+
 }
```
nchanged: Let's add one more
```
const require = function(str){ return str };
const hello = {
	foo  : {
		bar : {
			woo : () => return require("foobar")
			
		}
	}
}
	

module.exports = hello.foo.bar.woo()

```

and assertion

```
 const res = window.$fsx.r(0);
should(res).equal("foobar")
```

jpike88: this test fails in a weird way, it's out of the scope of what i'm fixing...

src/quantum/tests/GlobalVariable.test.ts
jpike88: screen shot 2018-04-27 at 9 03 02 pm

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -74,6 +74,63 @@ export class GlobalVariableTEst {
     }
 
 
+    public "Should replace require statement 2"() {
```
nchanged: nitpick: should NOT

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -74,6 +74,63 @@ export class GlobalVariableTEst {
...
+                    })]
+                }
+            }
+        ).simple().then(test => test.browser((window, env) => {
+            const app = env.getScript("app.js");
+            app.shouldFindString("isString(require)");
+        }));
+    }
+
+    public "Should replace require statement 3"() {
```
nchanged: Should NOT

src/quantum/tests/GlobalVariable.test.ts
```diff
@@ -140,4 +197,60 @@ export class GlobalVariableTEst {
...
+                    `
+                    },
+                    plugins: [QuantumPlugin({
+                        target: "browser"
+                    })]
+                }
+            }
+        ).simple().then(test => test.browser((window, env) => {
+            const app = env.getScript("app.js");
+            app.shouldNotFindString("require(2)");
```
nchanged: I think it should replace it? Because it was overridden earlier

nchanged:
@jpike88 Thank you very much for contributing and spending so much time polishing up this PR! ❤️

jpike88:
A project as awesome as fuse-box deserves everyone's love. I don't usually like to compliment things but seriously you deserve big kudos for this. I'll be around for other improvement and touch ups if you need the help!

PR #1235, opened 2018-04-28
fix: syntheticDefaultExportPolyfill Check if frozen before attempting to defineProperty.

### LINKED ISSUE #1224: mobx@4.2 triggers "Cannot redefine property: default" error
Hello,

I'm not sure if this is issue with mobx library or Fusebox.

I found that mobx-react@5 triggers "Cannot redefine property: default" error in browser console.
Previous versions of mobx-react work fine.

Example Repo: https://github.com/troynt/fuse-box-antd
Start using ```npm run watch```

Observe errors in console.

Errors disappear when mobx-react ```@observer``` decorator is removed. 
https://github.com/troynt/fuse-box-antd/blob/master/src/client/containers/App.tsx#L8

Mobx addition can be found in commit
https://github.com/troynt/fuse-box-antd/commit/2f2bfe6cebb70123d67997a97b7ecddd4443796e

Mobx code triggering error
https://github.com/mobxjs/mobx/blob/d17e73cba664842048008cefbbd2243a6db1af30/src/mobx.ts#L143-L161

Thanks,

Troy

troynt:
Fixes #1224

### DIFF
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -8,6 +8,7 @@
     /* @end */
 
     /* @if allowSyntheticDefaultImports */
+    // NOTE: Should match syntheticDefaultExportPolyfill in LoaderAPI.ts
     function syntheticDefaultExportPolyfill(input) {
         if( input === null ||
             ['function', 'object', 'array'].indexOf(typeof input) === -1 ||
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -8,10 +8,26 @@
     /* @end */
 
     /* @if allowSyntheticDefaultImports */
-    function syntheticDefaultExportPolyfill(input){
-        return input !== null && ['function', 'object', 'array']
-            .indexOf(typeof input) > -1 && input.default === undefined ?
-                 Object.isFrozen(input) ? input.default = input : Object.defineProperty(input, "default", {value : input, writable : true, enumerable : false}) : void 0;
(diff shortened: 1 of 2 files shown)


modules/fuse-box-responsive-api/index.js
```diff
@@ -8,10 +8,26 @@
     /* @end */
 
     /* @if allowSyntheticDefaultImports */
-    function syntheticDefaultExportPolyfill(input){
-        return input !== null && ['function', 'object', 'array']
-            .indexOf(typeof input) > -1 && input.default === undefined ?
-                 Object.isFrozen(input) ? input.default = input : Object.defineProperty(input, "default", {value : input, writable : true, enumerable : false}) : void 0;
+    function syntheticDefaultExportPolyfill(input) {
+        if( input === null ||
```
nchanged: shouldn't it be ` input !== null && ['function', 'object', 'array']` ?

modules/fuse-box-responsive-api/index.js
```diff
@@ -8,10 +8,26 @@
...
+            input.default = input;
+            return;
+        }
+
+        // free to define properties
+        Object.defineProperty(input, "default", {
+            value: input,
+            writable: true,
+            enumerable: false
+        });
```
nchanged: the same function should be in `src/loader/LoaderAPI.ts`

modules/fuse-box-responsive-api/index.js
nchanged: Maybe you could just add your solution to that ternary sausage?  ;-)

PR #1246, opened 2018-05-01
WebIndexPlugin additional attributes support in script tags

lanceschi:
Update WebIndexPlugin.ts in order to support the defer boolean attribute in script tags. Thanks

### DIFF
--- a/docs/plugins/html/WebIndexPlugin.md
+++ b/docs/plugins/html/WebIndexPlugin.md
@@ -34,6 +34,8 @@ fuse.plugin(
 | ` resolve `   | `resolve ?: {(output : UserOutput) : string}` Allows to completely override the output  |
 | `pre ` | `{ relType: 'fetch' | 'load' }`  Config object to either preload or prefetch the output javascript bundles in the head of the document |
 | ` appendBundles ` | Append $bundles to provided template |
+| ` async ` | Add the `async` attribute to javascript `<script>` tags (boolean) |
+| ` defer ` | Add the `defer` attribute to javascript `<script>` tags (boolean) |
 
 note: If you specify template and templateString then template will take precedent 
 
--- a/docs/plugins/html/WebIndexPlugin.md
+++ b/docs/plugins/html/WebIndexPlugin.md
@@ -34,10 +34,11 @@ fuse.plugin(
 | ` resolve `   | `resolve ?: {(output : UserOutput) : string}` Allows to completely override the output  |
 | `pre ` | `{ relType: 'fetch' | 'load' }`  Config object to either preload or prefetch the output javascript bundles in the head of the document |
 | ` appendBundles ` | Append $bundles to provided template |
-| ` async ` | Add the `async` attribute to javascript `<script>` tags (boolean) |
-| ` defer ` | Add the `defer` attribute to javascript `<script>` tags (boolean) |
+| ` scriptAttributes ` | Provide additional attribute(s) to `<script>` tags (type string) |
+
 
 note: If you specify template and templateString then template will take precedent 
+note: `scriptAttributes` option replace both previous `async` and `defer` options
 
 ### Resolve example
 `resolve` option allows you to completely override the path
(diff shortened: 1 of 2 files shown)


nchanged:
Can I ask you to update the doc?

lanceschi:
@Kompwu @nchanged doing right now

src/plugins/WebIndexPlugin.ts
```diff
@@ -86,7 +87,7 @@ $bundles
 
         let jsTags = this.opts.emitBundles
             ? this.opts.emitBundles(bundlePaths)
-            : bundlePaths.map(bundle => `<script ${this.opts.async ? 'async' : ''} type="text/javascript" src="${bundle}"></script>`).join('\n');
+            : bundlePaths.map(bundle => `<script ${this.opts.async ? 'async' : ''} ${this.opts.defer ? 'defer' : ''} type="text/javascript" src="${bundle}"></script>`).join('\n');
```
Kompwu: I think you should check if there is async or defer to avoid adding async plus defer and vice versa.

lanceschi: Ok, I got your point! Would you like me to keep both attributes as properties of the WebIndexPlugin or should I rewrite the as a generic js_additional_script_attributes (of type string)?

Kompwu: I think in that case, making `js_additional_script_attributes` would be more better.
It' d allow to do more things like for example add an `ID`

lanceschi:
@Kompwu I updated the docs too! ;)

docs/plugins/html/WebIndexPlugin.md
```diff
@@ -34,10 +34,11 @@ fuse.plugin(
 | ` resolve `   | `resolve ?: {(output : UserOutput) : string}` Allows to completely override the output  |
 | `pre ` | `{ relType: 'fetch' | 'load' }`  Config object to either preload or prefetch the output javascript bundles in the head of the document |
 | ` appendBundles ` | Append $bundles to provided template |
-| ` async ` | Add the `async` attribute to javascript `<script>` tags (boolean) |
-| ` defer ` | Add the `defer` attribute to javascript `<script>` tags (boolean) |
+| ` scriptAttributes ` | Provide additional attribute(s) to `<script>` tags (type string) |
```
Kompwu: To maintain consistency you should do `(string)` instead of `(type string)` :wink:

Kompwu:
Also, can you rename your PR? it doesn't really reflect what it' s about.

src/plugins/WebIndexPlugin.ts
```diff
@@ -18,7 +18,7 @@ export interface IndexPluginOptions {
     template?: string;
     templateString?: string;
     appendBundles?: boolean;
-    async?: boolean;
```
nchanged: Can we avoid breaking changes? Please keep it there and just update your object accordingly

lanceschi: @nchanged I got your point!

Is it ok to keep async property and add scriptAttributes as collector of other attributes?

Or is to be preferred a defer (boolean) property instead of scriptAttributes?

Let me know

src/plugins/WebIndexPlugin.ts
```diff
@@ -86,7 +87,7 @@ $bundles
 
         let jsTags = this.opts.emitBundles
             ? this.opts.emitBundles(bundlePaths)
-            : bundlePaths.map(bundle => `<script ${this.opts.async ? 'async' : ''} type="text/javascript" src="${bundle}"></script>`).join('\n');
+            : bundlePaths.map(bundle => `<script ${this.opts.async ? this.opts.async : ''} ${this.opts.scriptAttributes ? this.opts.scriptAttributes : ''} type="text/javascript" src="${bundle}"></script>`).join('\n');
```
nchanged: I believe it was `${this.opts.async ? 'async' : ''}`
sorry man for being picky ;-) But I don't want to have any breaking changes.

lanceschi: @nchanged no prob. I'm travelling these days. I missed that typo. Fixing and committing! Sorry and Thanks

nchanged:
❤️  thank you!

PR #1252, opened 2018-05-10
Update about-plugins.md

felixrabe:
Fix regex equivalent example.

### DIFF
--- a/docs/introduction/about-plugins.md
+++ b/docs/introduction/about-plugins.md
@@ -30,7 +30,7 @@ For example, the code snippet above is equivalent to:
 
 ```js
 plugins: [
-    [ /mypath\/**.css/, CSSPlugin() ]
+    [ /mypath\/\*\*\.css/, CSSPlugin() ]
 ]
 ```
 


docs/introduction/about-plugins.md
```diff
@@ -30,7 +30,7 @@ For example, the code snippet above is equivalent to:
 
 ```js
 plugins: [
-    [ /mypath\/**.css/, CSSPlugin() ]
+    [ /mypath\/\*\*\.css/, CSSPlugin() ]
```
nchanged: it's a valid RegExp I believe we should have `.*\.css`

felixrabe: Nope, the equivalent Regexp of `"mypath/**.css"` is ` /mypath\/\*\*\.css/`.

Example (see also https://www.google.com/search?q=js+escape+string+to+regex and https://stackoverflow.com/a/6969486):

    $ node
    > function escapeRegExp(str) {
    ...   return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
    ... }
    undefined
    > new RegExp(escapeRegExp("mypath/**.css"))
    /mypath\/\*\*\.css/

felixrabe: The characters '/' and '*' and '.' all need to be escaped.

Kompwu: I agree w/ @felixrabe .

nchanged: But ` /mypath\/\*\*\.css/` isn't a string. It's a REGEXP

felixrabe:
After thinking about it for a while, I came to the conclusion that maybe my understanding of the example might not be correct. I'm closing this and opening a new issue just to report the invalid RegExp, so @nchanged can implement a correct fix instead of this (probably incorrect) PR.

PR #1259, opened 2018-05-11
Copy readme so it's included in npm

Fjandin:
Copy README.md to dist folder so npm doesn't look so empty :P 

image


nchanged:
👍 thanks a lot!

PR #1267, opened 2018-05-15
Add ConsolidateWebIndexPlugin (WIP) [HELP WANTED]

### LINKED ISSUE #1244: Use the output of ConsolidatePlugin() as an index for WebIndexPlugin()
Everything I've tried has failed. :sob:
Can someone help me with this? :sos:

Kompwu:
Should fixes #1244 (When it's done)
Also, i have an issue when i inject it into plugins chain:
```
TypeError: ConsolidateWebIndexPlugin is not a function
```
(*very weird because i inject it right.*)

Here is my `fuse.js` for dev:
```js
const { FuseBox, ConsolidateWebIndexPlugin } = require("fuse-box")
const fuse = FuseBox.init({
    homeDir : "src",
    target : 'browser@es6',
    output : "dist/$name.js",
    plugins : [
      ConsolidateWebIndexPlugin({
          engine: 'pug',
          template: "./src/index.pug"
      })
    ]
})
fuse.dev() // launch http server
fuse.bundle("app").instructions(" > index.ts").hmr().watch()
fuse.run()
```

Also i'm looking for a way to use `file.context.homeDir` for `baseDir` & `includeDir`

### DIFF
--- a/src/plugins/ConsolidateWebIndexPlugin.ts
+++ b/src/plugins/ConsolidateWebIndexPlugin.ts
@@ -1,0 +1,131 @@
+import { Plugin } from "../core/WorkflowContext";
+import { BundleProducer } from "../core/BundleProducer";
+import * as fs from "fs";
+import { ensureAbsolutePath, joinFuseBoxPath } from "../Utils";
+import { UserOutput } from "../core/UserOutput";
+import * as path from "path";
+import * as Consolidate from "consolidate";
+
+export interface ConsolidateWebIndexPluginOptions {
+    engine: string;
+    extension?: string;
+    template?: string;
+    baseDir?: string;
+    includeDir?: string;
+    bundles?: string[];
+    path?: string;
+    target?: string;
+    appendBundles?: boolean;
+    async?: boolean;
+    scriptAttributes?: string;
+    pre?: { relType: 'fetch' | 'load' };
+    resolve?: { (output: UserOutput): string };
+    emitBundles?: (bundles: string[]) => string;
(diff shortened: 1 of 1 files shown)


Kompwu:
Ping @Fjandin @tomitrescak @jpike88 
I feel like there' s something you guys can do...

Kompwu:
@Fjandin It was my original idea, but after chatting with @nchanged 
He told me that even by modifying it was not possible and that it was better to make a plugin especially for that.
But if you find a way to do it, it might be cool.

Fjandin:
I can try later today

Kompwu:
@Fjandin Great! ping me when you've done.

nchanged:
I've probably misunderstood. What if we add `engine` option to WebIndexPlugin which will pick up the `Consolidate` module.
And as @Fjandin mentioned passing extra options 

```
WebIndexPlugin({
    template: 'src/index.pug',
    engine: 'pug',
    data: {some: 'data', to: 'use', in: 'template'}
})
```

So, long story short `engine` would force WebindexPlugin to process it using Consolidate

```
        let html = await Consolidate[this.opts.engine].render(this.opts.template, {
            cache: false,
            filename: 'base',
            basedir: this.opts.baseDir,
            includeDir: this.opts.includeDir
       })
```

That would be easier probably

nchanged:
@Kompwu You don't need ConslidatePlugin from FuseBox, you take only the part that transpiles it. You don't need `file` either, because you aren't bundling anything.

PR #1268, opened 2018-05-15
[WIP] WebIndexPlugin engine option

Fjandin:
POC to allow to use template engine via consolidate in WebIndexPlugin

example:
```js
WebIndexPlugin({
    template: 'src/index.ejs',
    engine: 'ejs'
}),
```

Will only work when you have `template` defined as i cannot figure out how to use `consolidate` with a string.

Are there maybe alternatives to `consolidate` that are syncronous and allow string input?

### DIFF
--- a/docs/development/dev-server.md
+++ b/docs/development/dev-server.md
@@ -31,6 +31,17 @@ fuse.dev({
 })
 ```
 
+## Fallback
+
+If you develop a SPA app you can set a fallback. All 404's will return the fallback file you specify. The file path is relative to your served root.
+
+```js
+fuse.dev({
+    root: 'public/build',
+    fallback: 'index.html'
+})
+```
+
 ## Port
 By default you will get `express` application running and a `socket` server (if [HMR](#hot-module-reload) is enabled) bound to port `4444`. To change the port provide `port` option.
 
--- a/docs/plugins/css/CSSModules.md
+++ b/docs/plugins/css/CSSModules.md
@@ -122,6 +122,14 @@ CSSModules({
 })
 ```
 
+### root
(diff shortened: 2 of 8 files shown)


nchanged:
@Fjandin I believe i've added it already ;-)
https://github.com/fuse-box/fuse-box/commit/e0cba107d73ea43aecf3429d3f899d5897e09bf9

Have I missed something?

Fjandin:
I think main difference is that in this sollution you will need to add `bundles`, `css` etc.. in the template language of your choosing and not as `$bundles`, `$css` etc...
example in ejs
```
    <div id="app"></div>
    <%-bundles%>
```
I kinda like that better tbh :P But that's just my preference

Fjandin:
Anywho. Closing this one

nchanged:
I think it should be agnostic to template engine, anyway generated script tags are the same 😀

nchanged:
@Fjandin If you want, we can pass `bundles` object to `locals` ?! But it should be simplified, as in

```json
[
   { path : "/bundles/app.js"}
]
```
Feel free to submit a PR on that, then we will cover all the scenarios

Fjandin:
@nchanged I'll create a PR with some suggestions 👍

Kompwu:
@Fjandin How's going with your PR ?

Fjandin:
Been swamped with work lately. Will make a PR later this week

Kompwu:
@Fjandin Hey, Any news ?

Kompwu:
@Fjandin ping

Fjandin:
Sorry, but still havent had any time.

PR #1269, opened 2018-05-17
 feat: Runtime Bundle Mapping from global or window object

DitlevHB:
🐢 Finally ready with a pull request. 
I hope it goes through! *fingers crossed*


nchanged:
@DitlevHB some tests have failed. I just want to ask you to write a few test cases
in this file

https://github.com/fuse-box/fuse-box/blob/a13815949a16c6a37a41d2264470732149d43b9e/src/tests/dynamic_imports_test/CodeSplittingFileIntegrity.test.ts#L29-L64

It's pretty simple, we need browser server and universal to check if the data is set correctly.

DitlevHB:
Hmm, not sure how it should check if the data is set correctly.
This is what i got so far:
```
 "Should if check browser data is it correctly"() {
        return FuseTestEnv.create(
            {
                project: {
                    files: {
                        "index.ts": `
                            export async function getRemoteFile() {
                                await import("./components/HomeComponent")
                            }
                        `,
                        "components/HomeComponent.ts": `
                            export function home(){ return "home"  }
                        `,

                    },
                    plugins: [
                        QuantumPlugin({
                            target: "browser",
                            runtimeBundleMapping: "newBundleMapping"
                        })
                    ]
                }
            }
        )
            .simple().then(test => test.browser((window, env) => {
                window['newBundleMapping'] = {
                    c: {
                        b: './asdasd',
                        s: './asdasdasd'},
                    i: {}
                };

                window.$fsx.r(0);

                
                env.scriptShouldExist(HOME_COMPONENT_SCRIPT);
               
                const appScript = env.getScript("app.js");
                
                appScript.shouldNotFindString('// default/components/HomeComponent.js');
                

                const homeScript = env.getScript(HOME_COMPONENT_SCRIPT)
                homeScript.shouldFindString('// default/components/HomeComponent.js');
            }));
    }

```

Wouldn't it make sense to try and import the homecomponent and see if it is successfull?

nchanged:
` window.$fsx.r(0);` that's the import of a first module

nchanged:
You can run a single test
`node test --file=CodeSplittingFileIntegrity.test.ts`

src/tests/dynamic_imports_test/CodeSplittingFileIntegrity.test.ts
```diff
@@ -705,6 +706,58 @@ export class CodeSplittingFileIntegrityTest {
...
+                            runtimeBundleMapping: "newBundleMapping"
+                        })
+                    ]
+                }
+            }
+        )
+            .simple().then(test => test.browser((window, env) => {
+                window['newBundleMapping'] = {
+                    c: {
+                        b: './asdasd',
```
nchanged: Let's give it real values, and see if it's possible to import without errors

DitlevHB: Well it doesn't fail with these 'non-real' values.
I wanted to make it fail first and then correct it with real values.

DitlevHB: It should fail on 
````
window.$fsx.r(0);
```
right?
Or am i missing the point

nchanged: can you hack into `FuseTestEnv` and add this `preloadScript` as an optional argument? That should do the trick

DitlevHB: So i successfully extended FuseTestEnv and added the preloading script.
I'm having issues validating, that the dynamic loading works properly.

```
 .simple().then(test => test.browser((window, env) => {
                var index = window.$fsx.r(0);
                index.getRemoteFile().then(result => {
                    should(result).equal("home");
                });
            }, bundleMappingScript));
```

Seems like async operations aren't possible within the first argument of FuseTestEnv.browser?

nchanged:
You need to return it ) if course

src/tests/dynamic_imports_test/CodeSplittingFileIntegrity.test.ts
nchanged: How is it going with it? Do you need help? You can just return it
```
 var index = window.$fsx.r(0);
                return index.getRemoteFile().then(result => {
                    should(result).equal("home");
                });
```

I can help you with this. Let's merge it already ;-)

DitlevHB: Hey I am really busy this week, I'll see what i can do. I might not be able to check in before friday.

DitlevHB:
Hey, have you seen I pushed the new tests?

nchanged:
@DitlevHB one thing we forgot to add is documentation ;-(

DitlevHB:
I'm busy with exams at the moment, but I'll add some documentation within a week or two 😉

PR #1276, opened 2018-05-27
Add .npmignore to keep everything smaller

Kompwu:
It’s not smart to include all of this into your npm package.
It’s a good practice to keep the size small.
Also, i'm not sure about `modules/` if it's needed or not, so i haven't touched it.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,6 @@
+.github
+docs
+src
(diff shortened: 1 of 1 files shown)


nchanged:
we are using a separate folder, we don't need .npmignore

PR #1281, opened 2018-05-30
Adding filesMapping option to CSSResourcePlugin

zouloux:
This option allow developer to create a manifest of generated resources files. Needed when loading CSS Resources from Javascript.
This fork manages watch loops to allow filesMapping function to create typescript or javascript files in homeDir folder.
filesMapping can also create a JSON file loaded at runtime for example.

Usage :
```
CSSResourcePlugin({

	dist: "resources/",

	// Called when a new set of resources is copied to the resources/ folder
	filesMapping: (files) =>
	{
		// Log copied files
		files.map( fileMapping =>
		{
			console.log(fileMapping.from, "has been copied to", fileMapping.to)
		});
	}
})
```

### DIFF
--- a/src/plugins/stylesheet/CSSResourcePlugin.ts
+++ b/src/plugins/stylesheet/CSSResourcePlugin.ts
@@ -10,6 +10,9 @@ const base64Img = require("base64-img");
 const IMG_CACHE = {};
 let resourceFolderChecked = false;
 
+// Needs a patch
+// @see : https://github.com/fuse-box/fuse-box/issues/1060
+// This fork is not about this
 const copyFile = (source, target) => {
     return new Promise((resolve, reject) => {
         fs.exists(source, (exists) => {
@@ -61,8 +64,9 @@ export interface CSSResourcePluginOptions {
     inline?: boolean;
     resolve?: (path: string) => any;
     macros?: any;
-    resolveMissing?: any,
-    useOriginalFilenames?: boolean
+    resolveMissing?: any;
+    useOriginalFilenames?: boolean;
+    filesMapping?: (files: {from: string, to: string}[]) => void
 }
 
 /**
@@ -78,6 +82,13 @@ export class CSSResourcePluginClass implements Plugin {
     public macros: any;
     public resolveMissingFn: any;
     public useOriginalFilenames: boolean = false;
+    
+    public files = {};
+    public copiedFiles: {from: string, to: string}[] = [];
+    public filesMapping: (files: {from: string, to: string}[]) => void;
+    public copiedFilesID: string = '';
+    public previousCopiedFilesID:string;
+    
     constructor(opts: CSSResourcePluginOptions = {}) {
         if (opts.dist) {
             this.distFolder = ensureDir(opts.dist);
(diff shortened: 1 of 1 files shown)


src/plugins/stylesheet/CSSResourcePlugin.ts
```diff
@@ -190,16 +203,54 @@ export class CSSResourcePluginClass implements Plugin {
...
+                    // We store this copied file source and destination path for the middleware
+                    this.copiedFiles.push({
+                        from: urlFile,
+                        to: newPath
+                    });
+
+                    // We also store a string which uniquely identify this array
+                    // To avoid watch loop
+                    this.copiedFilesID = this.copiedFiles.map(
+                        copiedFile => generateNewFileName( copiedFile.from )
```
nchanged: Why are we generating a new file name if it has been generated above? Besides, user might choose to use originalFileNames

src/plugins/stylesheet/CSSResourcePlugin.ts
```diff
@@ -190,16 +203,54 @@ export class CSSResourcePluginClass implements Plugin {
...
+        // Get home dir (src path)
+        const homeDir = producer.fuse.opts.homeDir;
+
+        // We store the copied file ID and continue only if it changed
+        // It allow us to avoid watch loop when fileMapping middleware is producing a source file
+        if (this.previousCopiedFilesID === this.copiedFilesID) return;
+        this.previousCopiedFilesID = this.copiedFilesID;
+
+        // Call middleware with copied files path
+        this.filesMapping(
```
nchanged: How about if filesMapping isn't defined, will it be still copied over?

src/plugins/stylesheet/CSSResourcePlugin.ts
zouloux: In fact we are using generateNewFileName to create a hash of the copiedFiles array. It is needed to identify when the list is updated. Maybe there is a better solution here.

Also, originalFileNames is not enough when we want generated file names and also JS aware of generated resources. This is why this option is needed.

src/plugins/stylesheet/CSSResourcePlugin.ts
zouloux: We are checking if this.fileMapping is defined from options at line 233. If it is not defined from options, file are still copied. Maybe this isn't clear enough ?

nchanged: Right, I just though that you changed the copying logic too. Can you update the doc too?

src/plugins/stylesheet/CSSResourcePlugin.ts
nchanged: How about checking for the option enabled? We don't need to generate those unless the option is set. Right?

zouloux: Right ! I'll create and test a cleaner pull request, without this copiedFileID logic. In fact, CSSResourcesPlugin only add resources to the list (removals are not detected), so only checking array length (and option enabled) should be enough.

src/plugins/stylesheet/CSSResourcePlugin.ts
zouloux: I'll try to do it but it will be in an another pull request I think.

Issue #1286, opened 2018-06-05
Package names that begin with capital letters can be defined, but not properly imported


dlannoye:
I have a repository that shows this issues here:
https://github.com/dlannoye/fuse-box-default-package

In fuse.js I am defining a package with a name that begins with an uppercase letter. This builds and bundles fine, but at load time I get an `Uncaught Package not found default` from the fuse-box generated import call. This is because the code in the loader doesn't treat the string beginning with a capital letters as a package name.

If I call `FuseBox.import("", { pkg: "SomeProject" });` the package loads just fine. 

I can understand this restriction since the [package.json](https://docs.npmjs.com/files/package.json) specs say not to use capital letters in a package name, but it would be nice if fuse-box handled this in a better way. Maybe through a build time error or a better exception message. 

Also the fuse-box [documentation for package name](https://fuse-box.org/page/configuration#package-name) uses invalid names for npm packages and doesn't mention these restrictions.

nchanged:
@dlannoye I see, the thing is that Quantum renders useless this package name option. So I would recommend removing that, and use Quantum plugin instead with `isolatedAPI`  or 'noConflictAPI'. If you want to expose it to `window` is `globals` option on FuseBox.init.

It will produce highly optimised no-conflict bundles .

dlannoye:
I had also noticed that Quantum avoids this issue and will probably be moving my development builds to use it as well. That said I have also opened a PR to update the documentation around package names: #1287

PR #1287, opened 2018-06-05
Add documentation around valid package names

dlannoye:
This updating the documentation so other don't run into the same issues I did with #1286

### DIFF
--- a/docs/development/configuration.md
+++ b/docs/development/configuration.md
@@ -204,7 +204,7 @@ note: TypeScript 2.4+ is required for custom transformers.
 Your default package name is `default`. You don't need to change it if you are not planning on having isolated bundles.
 Any bundle added as a script tag will share the `default` package, keep that in mind. If you want to release a package (say to NPM), you probably would want set a different name (to avoid scope collision)
 
-It's imperative to have a __unique name__ (matching an npm package) when publishing a bundle to NPM.
+It's imperative to have a __unique name__ (matching an npm package) when publishing a bundle to NPM. Your package name must start with either an '@' or a lowercase a-z character.
 
 
 ```js


docs/development/configuration.md
```diff
@@ -204,7 +204,7 @@ note: TypeScript 2.4+ is required for custom transformers.
 Your default package name is `default`. You don't need to change it if you are not planning on having isolated bundles.
 Any bundle added as a script tag will share the `default` package, keep that in mind. If you want to release a package (say to NPM), you probably would want set a different name (to avoid scope collision)
 
-It's imperative to have a __unique name__ (matching an npm package) when publishing a bundle to NPM.
+It's imperative to have a __unique name__ (matching an npm package) when publishing a bundle to NPM. Your package name must start with either an '@' or a lowercase a-z character.
```
nchanged: I don't think it can start with `@` since it's a symbol used to point to a scoped repository.

dlannoye: Take look at the [code in the LoaderAPI]( https://github.com/fuse-box/fuse-box/blob/9e0eb11d46f2c5c5d41299ff718af32d7c663d29/src/loader/LoaderAPI.ts#L127). This is the exact check it does:
```
    // https://www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes
    // basically lowcase alphabet starts with 97 ends with 122, and symbol @ is 64
    // which 2x faster than /^([@a-z].*)$/
    if (n >= 97 && n <= 122 || n === 64) {
```

dlannoye: I work with scoped packages and when I name them something like `@foo/bar` the generated folder structure in node_modules is something like:
```
`@foo`
   -> `bar`
```
And you can definitely import from them by doing `import "@foo/bar"` there is no need for an extra '@'.

Am I misunderstanding your concern? 

Also I can confirm fuse-box works fine with package names beginning with a scope defined in the config. Take a look at this sample:
https://github.com/dlannoye/fuse-box-default-package/tree/scoped-package

nchanged: Right, I would recommend you to avoid this package option and use Quantum instead tho ;-)

PR #1300, opened 2018-06-25
fix: Target 'npm' more like 'universal' (nor 'browser' nor 'server')

eddow:
`npm` target shouldn't be neither "browser" neither "server" as there are browser and server npm packages. It should more be like universal and determine every time if it is in a browser or not. This is especially important for Vue packages to allow SSR

Note: I noticed the problem when my CSS was not included in my browser-targeted npm package as `FuseBox.isBrowser` was `false`.

### DIFF
--- a/src/quantum/plugin/modifications/EnvironmentConditionModification.ts
+++ b/src/quantum/plugin/modifications/EnvironmentConditionModification.ts
@@ -9,7 +9,7 @@ export class EnvironmentConditionModification {
 
         return each(file.fuseboxIsEnvConditions, (replacable: ReplaceableBlock) => {
 
-            if (core.opts.isTargetUniveral()) {
+            if (core.opts.isTargetUniveral() || core.opts.isTargetNpm()) {
                 if (replacable.identifier === "isServer") {
                     replacable.setFunctionName(`${core.opts.quantumVariableName}.cs`);
                 }
--- a/src/quantum/plugin/QuantumOptions.ts
+++ b/src/quantum/plugin/QuantumOptions.ts
@@ -339,7 +339,7 @@ export class QuantumOptions {
     }
 
     public isTargetServer() {
-        return this.optsTarget === "server" || this.optsTarget === "electron" || this.optsTarget === "npm";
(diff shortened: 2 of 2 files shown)


src/quantum/plugin/modifications/EnvironmentConditionModification.ts
```diff
@@ -9,7 +9,7 @@ export class EnvironmentConditionModification {
 
         return each(file.fuseboxIsEnvConditions, (replacable: ReplaceableBlock) => {
 
-            if (core.opts.isTargetUniveral()) {
+            if (core.opts.isTargetUniveral() || core.opts.isTargetNpm()) {
```
nchanged: I a not sure it's required. Target npm make a server bundle anyway. Once bundled an appropriate target will be set. So say, you publish with `npm` target but once bundled, you can choose `server` or `browser` or `universal`.

eddow:
"Target npm make a server bundle anyway"
That is where I disagree. I am trying to make a browser NPM package and it cannot deploy CSS because of that one line : it is considered as server while I am trying to make a browser-side npm package

src/quantum/plugin/modifications/EnvironmentConditionModification.ts
eddow: "Target npm make a server bundle anyway"
That is where I disagree. I am trying to make a browser NPM package and it cannot deploy CSS because of that one line : it is considered as server while I am trying to make a browser-side npm package.

The produced NPM package produces this:
```
var runningInBrowser = false || 'npm' === 'electron';
```

while it should check for the presence of the `window` global variable instead of declaring it `false`. And this is what this PR is about.

I mean, I don't get where this idea that "npm = server" comes from. jQuery for exemple is on npm for G. sake, saying that jQuery is a server-side library and avoiding it to touch CSS is non-sense.

This PR brings no breaking, it just checks each time whether it is browser or server-side instead of blindly considering it is server-side. On the other hands, it allows browser-side npm packages to be bundled.

nchanged:
I am closing it now due to inability to understand the issue

PR #1306, opened 2018-06-29
Update WebIndexPlugin to preload multiple types.

JohnPittman:
This gives the ability to preload multiple tag types and override duplicate injections.

I was one of the original people to request the ability to automate preloading resources within the index.html via link tags. I noticed that y'all implemented it for script files but the request was for CSS files so I decided to take a stab at adding it in for you. I did this within the raw editor on github so nothing has been tested but it should work and give you a sense of what it should be. 

Let me know what you think. Thanks!

### DIFF
--- a/src/plugins/WebIndexPlugin.ts
+++ b/src/plugins/WebIndexPlugin.ts
@@ -102,34 +102,85 @@ $bundles
             }
         }
 
-        let jsTags = this.opts.emitBundles
-            ? this.opts.emitBundles(bundlePaths)
-            : bundlePaths.map(bundle => `<script ${this.opts.async ? 'async' : ''} ${this.opts.scriptAttributes ? this.opts.scriptAttributes : ''} type="text/javascript" src="${bundle}"></script>`).join('\n');
+                // Generate preload and prefetch resource tags by tag type.
+        // ex. pre: {relType: 'load', tagTypes: ['style', 'script']}
+        let preloadTags = '';
+        let preloadCSS = false;
+        let preloadJS = false;
 
-        let preloadTags;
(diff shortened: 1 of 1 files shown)


nchanged:
@JohnPittman thanks,
Can you give me some use cases and/or examples on how to use it?

JohnPittman:
I really only care about the link tags getting preloaded to not block script tags from getting loaded so the simplest example is that I would want my base styles.css file preloaded. 

Say I expect:  <link rel="preload" href="style.css" as="style">

Plugin config:

WebIndexPlugin({
            appendBundles: true,
            pre: {
                relType: 'load',
                tagTypes: ['style']
            },
            target: 'index.html',
            template: 'src/templates/index.html',
            title: 'React Example'
        }),

JohnPittman:
Personally I think you would be better off controlling each type individually via 'appendBundles' : { preload: true } and add 'appendCSS' etc. instead of relying on the inject option from the CSSPlugin.

nchanged:
Right, well there are quite an amount of changes. We need tests, I don't want to have any regressions.

JohnPittman:
Also need to add 'defer' option to script tags where a script can be 'async' or 'defer'.  both are very relevant.

JohnPittman:
I am going to end up writing my own WebIndexPlugin for a much robust usage. Thanks anyway.

nchanged:
Feel free to share it with us! 😃

JohnPittman:
Version 1 that should solve 95% of needs:

https://github.com/chickendinosaur/fuse-box-web-index-plugin

chickendinosaur:
This is my main account so adding comment to chain.

Issue #904, opened 2017-10-27
Fusebox not loading JSON within a node module


calebboyd:
JSON files within a module don't seem to resolve...
Instead of throwing (#902) the value is `null`

See example repo:
https://github.com/calebboyd/fusebox-json

calebboyd:
This is clearly a bug.

phamtm:
@calebboyd shouldn't you require `test-module` using `var test = require("./test-module")`, notice the `./` there? 

I think without specifying the relative path then `fusebox` may think it is a third party module. Refers to https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders . So the problem should be that `fusebox` does not throws an error/warning when some module is not found.

nchanged:
@phamtm you are right
@calebboyd you either use `aliases` or `modulesFolder`

calebboyd:
@phamtm yeah, typo, corrected the issue still persists, Looks like the latest (2.5.0-beta.1) also introduces #791 again.

Its probably time to add typescript as a peer dependency... /cc @devmondo

calebboyd:
@phamtm Actually it wasn't a typo...(just back from holiday -- forgot how It was setup) 

Its copied into node_modules - see https://github.com/calebboyd/fusebox-json/blob/master/package.json#L6

I did that to illustrate it with behavior with a node module. But it looks like its present as a non-node module as well.

dattanchu:
I am running into the same problem. Fuse will not resolve .json files shipped in a package installed.

The fix for the equivalent problem in Webpack is to add .json to resolve.extensions

```
{
  resolve: {
    extensions: [".js", ".json", ...]
  }
}
```

soanaphikeste:
The solution/workaround seems to be the JSONPlugin: https://fuse-box.org/page/json-plugin#usage

When I include it in my config, everything works as expected.

grossbart:
Node modules `require`ing JSON config files no longer break with the JSONPlugin, however it breaks my requirement of loading JSON asynchronously using the CopyPlugin. At the moment it doesn't seem possible to have both because JSONPlugin doesn't support any options to include/exclude files and folders.

nchanged:
You can a specific path matching your files, first argument in the chain

grossbart:
I see, thank you for the hint @nchanged. Something like this is what I'm using now:

```
plugins: [
    [ "node_modules/**.json", JSONPlugin() ],
    CopyPlugin(…),
    …
]
[...]

PR #1310, opened 2018-07-03
#904 allow `require('config')` to load json

nalbion:


nalbion:
I'll admit, it's a bit ugly...

`ExtensionOverrides.setOverrideFileInfo()` returns early if for files in `node_modules`, and it also does not update `file.relativePath`, so the JSON plugin would not fire anyway.

Because this is a language feature, this should all work out-of-the-box without having to explicitly set `extensionOverrides: ['.json']`

nchanged:
What's the actual purpose of it?

nalbion:
@nchanged I'm working (trying to) with `googleapi`. Their code includes 

    var config = require('./some-config-file');

Fusebox currently tries to load `some-config-file.js`, but it needs to load `some-config-file.json`

nalbion:
@nchanged this should resolve #904

nchanged:
Oh I see, In this case we should fix

nchanged:
@nalbion 
How about aliases?

```
alias : {
  "some-config-file" : "some-config-file.json" 
}
```
That should solve your particular issue. Of course, we need to fix it on FuseBox level, but that would  require QuantumPlugin fix as well

Kompwu:
@nchanged Can we close it now?

nchanged:
Closing this for now.

PR #1312, opened 2018-07-04
Simplify pre option in WebIndexPlugin and add documentation

chrstntdd:
I know this is a breaking API change, but I figured the feature I established didn't offer an intuitive configuration. Instead of specifying as an object, a string enum is of 'fetch' or 'load' is required. There were also some internal changes including:

*  Use the more expressive `!string.includes()` instead of `string.indexOf() === -1`
* Prefer `const` over `let` where possible
* Unify the use of single quotes for strings
* Sort object properties alphabetically
* Improve spacing consistency

Additionally, the documentation for WebIndexPlugin has been updated to reflect the API changes as well as documenting all the options that the plugin can accept. I did my best to document what I could and would appreciate some reviews of my explanations so that this plugin can be more easily understood by newcomers.

### DIFF
--- a/docs/plugins/html/WebIndexPlugin.md
+++ b/docs/plugins/html/WebIndexPlugin.md
@@ -10,7 +10,7 @@ Generates a HTML file once a producer's job is completed
 Import from FuseBox
 
 ```js
-const { WebIndexPlugin } = require("fuse-box");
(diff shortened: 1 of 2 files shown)


chrstntdd:
oh wow, I don't think the branch I worked on was up to date 🤦‍♂️

PR #1313, opened 2018-07-04
Simplify pre option in WebIndexPlugin and add documentation

chrstntdd:
Re-opening due to an out of date branch that I was working on before.

I know this is a breaking API change, but I figured the feature I established didn't offer an intuitive configuration. Instead of specifying as an object, a string enum is of 'fetch' or 'load' is required. There were also some internal changes   ##including:

*  Use the more expressive `!string.includes()` instead of `string.indexOf() === -1`
* Prefer `const` over `let` where possible
* Unify the use of single quotes for strings
* Sort object properties alphabetically
* Improve spacing consistency

Additionally, the documentation for WebIndexPlugin has been updated to reflect the small API change as well as documenting all the options that the plugin can accept. I did my best to document what I could and would appreciate some reviews of my explanations so that this plugin can be more easily understood by newcomers.


chrstntdd:
As I was looking over this file, I was noticing that there's some room to simplify the configuration and code here. For instance, there are now 2 ways to set the `async` attribute to a script tag and 3 ways to set some meta tags. Does anyone else think this is worth the time?

nchanged:
@chrstntdd So what's the breaking change exactly? 
Could you change single quotes to double? I prefer them everywhere.

chrstntdd:
@nchanged the breaking change is the switch from setting option to create `<link rel="preload" as="script" href="module.js" />` tags from being:

```js
WebIndexPlugin({
  pre: { relType: 'load' }
})
```

to

```js
WebIndexPlugin({
  pre: 'load'
})
```

Re: single quotes. Is there a reason this project hasn't adopted an automatic formatter that would enforce this kind of consistency on a pre-commit hook?

Also, I think I'm missing some error handling for the `pre` option. Will add that in another commit when I address the quote consistency.

nchanged:
@chrstntdd let's keep it compatible with the old API (we need to avoid breaking changes) but use the new API (and update the doc)

src/plugins/WebIndexPlugin.ts
```diff
@@ -1,78 +1,86 @@
...
     charset?: string;
     description?: string;
+    emitBundles?: (bundles: string[]) => string;
+    engine?: string;
     keywords?: string;
-    author?: string;
-    bundles?: string[];
+    locals?: { [key: string]: any }
     path?: string;
+    pre?: string;
```
Kompwu: Please keep it backward compatible.

nchanged:
One thing, can I ask you to fix typescript errors? Basically `typeof this.opts.pre === "object"`

Kompwu:
@nchanged I can do it if you want :)

nchanged:
@Kompwu you sure you can do it? =) You would need to clone the project and run `gulp dist` to verify the fix.

PR #1338, opened 2018-08-03
fix: log.showBundledFiles ignored when log.enabled value missing

sod:
Tried to set `log.showBundledFiles = false`, but it had no effect. Had to look in the source code why. 

`doLog` is always just set back to default, so this PR changes to don't set it at all. And always apply `log.showBundlesFiles`, if found.

Fixes following configs:

```js
FuseBox.init({
    log: {
        showBundledFiles: false, // <- was ignored
    },
});

FuseBox.init({
    log: {
        enabled: false, // <- was ignored
        showBundledFiles: false, // <- was ignored
    },
});
```

Those cases now work plus cases that already worked like:

```js
FuseBox.init({
    log: {
        enabled: true,
        showBundledFiles: false,
    },
});

FuseBox.init({
    log: true,
});

FuseBox.init({
    log: false,
});

FuseBox.init({
    // defaults to true if not set
});
```

@nchanged

### DIFF
--- a/src/core/FuseBox.ts
+++ b/src/core/FuseBox.ts
@@ -228,21 +228,13 @@ export class FuseBox {
             this.context.filterFile = opts.filterFile;
         }
 
-        if (opts.log) {
-            if (typeof opts.log === "boolean") {
-                this.context.doLog = opts.log;
-            }
-
-            if (typeof opts.log === "object" && opts.log.enabled) {
-                this.context.doLog = opts.log.enabled;
-                this.context.log.printLog = opts.log.enabled;
-                this.context.log.showBundledFiles = opts.log.showBundledFiles;
-            }
+        if (typeof opts.log === "boolean") {
+            this.context.log.printLog = opts.log;
         }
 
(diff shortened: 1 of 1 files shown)


src/core/FuseBox.ts
```diff
@@ -228,21 +228,13 @@ export class FuseBox {
...
-                this.context.doLog = opts.log.enabled;
-                this.context.log.printLog = opts.log.enabled;
-                this.context.log.showBundledFiles = opts.log.showBundledFiles;
-            }
+        if (typeof opts.log === "boolean") {
+            this.context.log.printLog = opts.log;
         }
 
-        if (opts.log !== undefined) {
-            this.context.doLog = true;
```
nchanged: What happened to doLog? I see it's been removed.

sod: It's on true already https://github.com/sod/fuse-box/blob/master/src/core/WorkflowContext.ts#L155

And was never really user overwritable because this line just set it back to true.

PR #1344, opened 2018-08-13
docs(CONTRIBUTING): Refined contrubuting guidelines.

merlinnot:
Hey,

I've tried to refine contributing guidelines a bit, hope you'll like the changes. I've also moved it to the root of the project so it's not GitHub-specific and it's easier to find. I've found some useful notes about local development there, but at first I didn't really think to look for these in `.github` directory.

Maybe we could add a link in the Readme?

Cheers


nchanged:
That's brilliant! Thanks so much for helping out! I really appreciate that!

Issue #1184, opened 2018-04-07
Quantum bug


leon-andria:
Hi, I have a very annoying issue with quantum, 
This is the index.ts code I have 
```js
(async () => {
    const moment = await import("moment")
    console.log(moment);
 })();
```
Now with this simple standard config 
```js
const sharedFuseConfig = {
    homeDir: `src`,
    cache: true,
    modulesFolder: `${__dirname}/node_modules/`,
    plugins: [
      QuantumPlugin({
            target : 'browser',
            bakeApiIntoBundle : 'app',
            manifest : true,
            containedAPI : false,
            removeExportsInterop : false,
            replaceTypeOf:false,
            removeUseStrict : false,
            ensureES5 : false
        })
    ],
    output: `./dist/$name.js`
};
```
Now when I do node fuse, everything is fine and fuse creates a split bundle for `momentjs` but now if I modify my `index.ts` and save it fuse watcher kicks again but now I get an empty `momentjs` splitted file like this:
```js
(function($fsx){
})($fsx);
```
 The only way around this is to stop the fuse process everytime I change my local code and restart so quantum will create a none empty split `momentjs` file.

So to explain better, what is happening is that with quantum enabled, it works and kicks  but it creates empty split files and you have to restart fuse process in order to avoid this issue. Which is something I don’t want at all, because for me I want to see my production result immediately and I don’t want to go through the dev vs production process. That is why I enabled quantum by default.

nchanged:
I get it, will check it out next week, thanks!

leon-andria:
Keep me update on your findings. This bug is really a blocking point.

nchanged:
I will get to it when I will have time.

leon-andria:
Thanks for your support!
By the way, I debug it and I found out where is the bug. It takes 20 min to figure it out but the bug is special on this line: https://github.com/fuse-box/fuse-box/blob/master/src/quantum/plugin/QuantumCore.ts#L167
Specially this line is so confusing but I guess, you know what you wrote. To solve the bug, I have to add the following line after it:
```js
this.producerAbstraction.bundleAbstractions.get(key).packageAbstractions = bundleAbstraction.packageAbstractions;
```
The problem is that after the first bundle, packageAbstractions from bits are empty. So, if you want the render to write content, https://github.com/fuse-box/fuse-box/blob/master/src/quantum/plugin/QuantumCore.ts#L267. You have to fullfil packageAbstractions.

[...]

PR #1350, opened 2018-08-15
fix(Quantum): Prevent Quantum from creating empty chunks

### LINKED ISSUE #1297: Tiny chunk: 5 bytes
In my production build, a chunk of size 5 bytes is created.

Here is a content of the file:
```js
$fsx;
```

All of the logs for this chunk:
```
  → Create split bundle 3b2ebcd7 with entry point @firebase/firestore/dist/index.cjs.js

  → Render bundle 3b2ebcd7

  → Uglifying bundles/3b2ebcd7...
  → Using uglify-es because the target is greater than ES5 or es6 option is set
  → Done uglifying bundles/3b2ebcd7
  size:  5 Bytes, 25 Bytes (gzipped)
```

nchanged:
fixes #1297 #1184

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,5 +123,5 @@
     "test": "node test.js"
   },
   "typings": "index.d.ts",
-  "version": "3.5.0-next.8"
+  "version": "3.5.0-next.9"
 }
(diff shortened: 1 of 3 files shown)


src/tests/stubs/FuseTestEnv.ts
```diff
@@ -151,6 +151,10 @@ export class FuseTestEnv {
 		return this.scripts.get(name);
 	}
 
+	public getScripts() {
```
merlinnot: Why don't you just make a getter?

nchanged: java background ;-))

merlinnot: It's a public field anyways, so actually... what's the point of this method? :)

PR #1355, opened 2018-08-17
[Issue-1308] Implement remote script .css/.js order loading in parallel.

johnpittman:
Chunk files that have a corresponding .css file loaded together will download in parallel but wait for the browser to execute the CSS first before running the chunk.

### DIFF
--- a/modules/fuse-box-responsive-api/index.js
+++ b/modules/fuse-box-responsive-api/index.js
@@ -130,45 +130,51 @@
 	}
 	/* @end */
 
-	/* @if loadRemoteScript  */
-	function loadRemoteScript(url, isCSS) {
-		/* @if server */
-		return Promise.resolve();
-		/* @end */
-
-		/* @if browser || universal */
-		return Promise.resolve().then(function() {
-			/* @if universal */
-			if (!isBrowser) {
-				return;
-			}
-			/* @end */
-
-			var d = document;
-			var head = d.getElementsByTagName("head")[0];
-			var target;
-			/* @if cssLoader */
-			if (isCSS) {
-				target = d.createElement("link");
-				target.rel = "stylesheet";
-				target.type = "text/css";
-				target.href = url;
-			} else {
-				/* @end */
-
-				target = d.createElement("script");
-				target.type = "text/javascript";
-				target.src = url;
-				target.async = true;
-
-				/* @if cssLoader */
-			}
-			/* @end */
-			head.insertBefore(target, head.firstChild);
-		});
-		/* @end */
-	}
+/* @if loadRemoteScript  */
+function loadRemoteScript(url, isCSS) {
+	/* @if server */
+	return Promise.resolve();
 	/* @end */
+
+	/* @if browser || universal */
+	return new Promise(function(resolve, reject) {
+		/* @if universal */
+		if (!isBrowser) {
+			return resolve();
+		}
+		/* @end */
+
+		var d = document;
+		var head = d.getElementsByTagName("head")[0];
+		var target;
+		/* @if cssLoader */
+		if (isCSS) {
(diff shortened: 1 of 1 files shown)


modules/fuse-box-responsive-api/index.js
```diff
@@ -301,28 +330,15 @@
 				/* @end */
 
 				if (data[2] && data[2].css === true) {
-					loadRemoteScript(path + data[2].name, true);
+					Promise.all([loadRemoteScript(path + data[2].name, true), new Promise((resolve, reject) => {
+						loadScript(path, data, $cache, id, resolve, reject)
+					})
+				]).then((values) => {
```
nchanged: Can you use `function` here? To keep the IE11 from breaking?

modules/fuse-box-responsive-api/index.js
```diff
@@ -130,45 +130,51 @@
...
-				target.type = "text/css";
-				target.href = url;
-			} else {
-				/* @end */
+			target = d.createElement("script");
+			target.type = "text/javascript";
+			target.onload = resolve;
+			target.onerror = reject;
+			// Should this be defer to keep the order of script execution? Script still download in parallel but load in order of other defers.
+			// target.defer = true;
```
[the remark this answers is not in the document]
johnpittman: You are correct about defer. Async is the correct choice for chunks although that section of code never gets used anyway.

Not sure about the git hook. I just ```git add --all -> git commit -m"msg" -> git push```

modules/fuse-box-responsive-api/index.js
johnpittman: Use whatever you need for it to work. not sure for ```function``` is. Does this code get babel'd because that would leave it up to the user to choose what they support via babel-env?

nchanged: Not I does't. It's your PR, you should fix it ;-)
```
then(function(values) => {
```

Simple is that.

modules/fuse-box-responsive-api/index.js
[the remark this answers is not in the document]
nchanged: So we are good there ;-)

modules/fuse-box-responsive-api/index.js
johnpittman: Didn't know you were talking about the arrow function. Sorry.

johnpittman:
Don't arrow functions come with the same compatibility as Promises?

modules/fuse-box-responsive-api/index.js
```diff
@@ -301,28 +328,15 @@
 				/* @end */
 
 				if (data[2] && data[2].css === true) {
-					loadRemoteScript(path + data[2].name, true);
+					Promise.all([loadRemoteScript(path + data[2].name, true), new Promise(function(resolve, reject) {
+						loadScript(path, data, $cache, id, resolve, reject);
+					});
+				]).then((values) => {
```
nchanged: How about here?
We have Promises polyfill, but not that.

nchanged:
There is a test failing for some reason

nchanged:
@johnpittman  there is some serious syntax error at `modules/fuse-box-responsive-api/index.js`

modules/fuse-box-responsive-api/index.js
```diff
@@ -301,28 +328,15 @@
 				/* @end */
 
 				if (data[2] && data[2].css === true) {
-					loadRemoteScript(path + data[2].name, true);
+					Promise.all([loadRemoteScript(path + data[2].name, true), new Promise(function(resolve, reject) {
+						loadScript(path, data, $cache, id, resolve, reject);
+					});
```
nchanged: Please remove the semicolon.

PR #1356, opened 2018-08-17
build(devDependencies): Use devDependencies instead of a custom Gulp task

merlinnot:
Using devDependencies enables more npm utilities to be used, like `npm audit`, `npm outdated`. It
also adds entries to the package-lock.json, which is important for the performance of a CI/CD
systems.


nchanged:
@merlinnot let's discuss this thingy. We need to get rid of some old tests

gulpfile.js
```diff
@@ -352,46 +352,3 @@ gulp.task("dev", () => {
...
-		"coffee-script",
-		"@types/node",
-		"vue-template-compiler",
-		"vue-template-es2015-compiler",
-		"vue",
-		"jwt-decode",
-		"vue-server-renderer",
-		"vue-hot-reload-api",
-		"vue-class-component",
-		"rollup",
```
nchanged: Rollup we don't need 100%

gulpfile.js
```diff
@@ -352,46 +352,3 @@ gulp.task("dev", () => {
...
-		"buble",
-		"consolidate",
-		"pug",
-		"tslint",
```
nchanged: Let's remove tslint for now. We will add it later

gulpfile.js
```diff
@@ -352,46 +352,3 @@ gulp.task("dev", () => {
...
-		"@angular/core",
-		"stylus",
-		"less",
-		"postcss",
-		"postcss-selector-parser",
-		"marked",
-		"node-sass",
-		"uglify-js",
-		"uglify-es",
-		"source-map@0.5.7",
```
nchanged: Not sure if it's needed here.

gulpfile.js
```diff
@@ -352,46 +352,3 @@ gulp.task("dev", () => {
 		},
 	);
 });
-gulp.task("installDevDeps", function(done) {
-	var deps = [
-		"babel-core",
-		"babel-generator",
```
nchanged: let's remove babel-core and babel-generator tests. They are useless

gulpfile.js
```diff
@@ -352,46 +352,3 @@ gulp.task("dev", () => {
...
-		"babel-preset-latest",
-		"babel-plugin-transform-es2015-modules-commonjs",
-		"babylon",
-		"cheerio",
-		"@angular/core",
```
nchanged: Questionable. I think we should remove it too

gulpfile.js
```diff
@@ -352,46 +352,3 @@ gulp.task("dev", () => {
...
-		"node-sass",
-		"uglify-js",
-		"uglify-es",
-		"source-map@0.5.7",
-		"coffee-script",
-		"@types/node",
-		"vue-template-compiler",
-		"vue-template-es2015-compiler",
-		"vue",
-		"jwt-decode",
```
nchanged: Do we really need this one. Hm. ?!

gulpfile.js
merlinnot: I've updated it and tested. There is a 1.0.0-rc.2 release, I hope it will stabilise soon.

nchanged: Actually, we don't have tests for it... So you can remove it

merlinnot: It's a dependency, not a devDependency. Are you sure? I'm fairly certain it should stay as is.

merlinnot:
@nchanged can you take another look?

gulpfile.js
nchanged: oh yeah you are right. it's just not used. It's for those who don't use any transpilers.

PR #1358, opened 2018-08-17
chore(bin/fusebox.js): Remove bin/fusebox.js as it can be specified as a dependency

merlinnot:
fuse-box is now a dev dependency, so the version can be changed freely


nchanged:
@merlinnot the whole point of bin/fusebox.js is to test fusebox. It won't make sense if a buggy version tests the same version

merlinnot:
A stable version, specified in the `package.json` will be used for testing.

Issue #997, opened 2017-12-29
CSSModules generates different hash when creating two bundles at once


dzek69:
This is a summary of what I've reported on gitter.

Comparing to 2.5.0 beta.1 current version (3.0.0 beta.44) Fusebox now forces `generateScopedName` option being passed to `postcss-modules` by doing this:
```javascript
generateScopedName: this.scopedName ? this.scopedName : '_[local]___[hash:base64:5]'
```

And there are two issues with that:
1) I may actually want to leave it undefined, so `postcss-modules` will keep its default behavior

2) when `generateScopedName` is a string and uses `hash` - the hash will be calculated different if two bundle are created in the same process

Why?
`postcss-modules` are using `generic-names` to generate the hash and uses `filename` for that.

But files seems to be passed to `postcss` as content strings not filenames:
```javascript
return postcss([
    require('postcss-modules')({ /* options */ })
]).process(file.contents, {})
    .then(result => {
        file.contents = result.css;
        return resolve();
    });
```
This causes filenames inside `generic-names` to be visible as: `<input css 1>` `<input css 2>` ... `<input css 13>` and so on.

So when I'm making first bundle - my file will get `<input css 1>` as a base for hash generating, but when second bundle is made - it will pass the same css contents again and will result with `<input css 14>` as filename.

So the hashes will be different.

nchanged:
@dzek69 could you suggest a fix?

henrinormak:
@nchanged what if the postcss usage in CSSModules plugin would feed in the file path as the from option?

```
return postcss([
    require('postcss-modules')({ from: file.absPath })
]).process(file.contents, {})...
```

In my local test, this seems to produce stable hashes for same files/classes.

nchanged:
We can try that! Can you open a PR?

PR #1377, opened 2018-09-07
fix: use correct CSS file path

henrinormak:
Possible fix for #997.

Not sure which test case I should update, didn't see any covering CSSModules plugin.


henrinormak:
Any ideas when this might get a review? Really need the fix for a deadline and wondering if I should publish my fork or not.

PR #1411, opened 2018-10-17
Extends

### LINKED ISSUE #590: fuse-box ignoring 'extends' in tsconfig.json
Hi, i just noticed that "when a tsconfig file uses "extends" it gets ignored when using fuse-box, yet works correctly when using tsc directly.

For example, say I have these two files:
../tsconfig.json, with compiler options such as "jsx": "React"
./tsconfig.json, with "extends": "../tsconfig.json"

If I run tsc againt .tsx files it works correctly, since ./tsconfig correctly extends the base one.
Fuse-box fails with 
```
  → Typescript config:  \tsconfig.json
Acorn error: Unexpected token (75:51)
```

but if I remove the extends and move the "jsx": "React" to the local file then it works correctly.

unional:
closes #590 

I'm not familiar with the test system so I don't know how to write a test for it.

I have tested this in my own repo usage.

### DIFF
--- a/src/core/TypescriptConfig.ts
+++ b/src/core/TypescriptConfig.ts
@@ -110,7 +110,6 @@ export class TypescriptConfig {
 				const configFileRelPath = this.configFile.replace(this.context.appRoot, "");
 				this.context.log.echoInfo(`Typescript config file:  ${configFileRelPath}`);
 				configFileFound = true;
-				console.log(this.configFile)
 				const res = ts.readConfigFile(this.configFile, ts.sys.readFile);
 				config = { ...res.config, compilerOptions: ts.parseJsonConfigFileContent(res.config, ts.sys, this.context.appRoot).options }
 				if (res.error) {
--- a/src/core/TypescriptConfig.ts
+++ b/src/core/TypescriptConfig.ts
@@ -64,6 +64,7 @@ export class TypescriptConfig {
 	private initializeConfig() {
 		const compilerOptions = this.config.compilerOptions;
 		compilerOptions.jsx = "react";
+		compilerOptions.importHelpers = true;
 		compilerOptions.emitDecoratorMetadata = true;
 		compilerOptions.experimentalDecorators = true;
 		const targetFile = path.join(this.context.homeDir, "tsconfig.json");
--- a/src/core/TypescriptConfig.ts
+++ b/src/core/TypescriptConfig.ts
@@ -14,7 +14,7 @@ export class TypescriptConfig {
 	private customTsConfig: string;
 	private configFile: string;
 
-	constructor(public context: WorkFlowContext) {}
+	constructor(public context: WorkFlowContext) { }
 
 	public getConfig() {
 		this.read();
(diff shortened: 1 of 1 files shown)


src/core/TypescriptConfig.ts
```diff
@@ -110,8 +110,8 @@ export class TypescriptConfig {
 				const configFileRelPath = this.configFile.replace(this.context.appRoot, "");
 				this.context.log.echoInfo(`Typescript config file:  ${configFileRelPath}`);
 				configFileFound = true;
-				const res = ts.readConfigFile(this.configFile, p => fs.readFileSync(p).toString());
-				config = res.config;
+				const res = ts.readConfigFile(this.configFile, ts.sys.readFile);
+				config = { ...res.config, compilerOptions: ts.parseJsonConfigFileContent(res.config, ts.sys, this.context.appRoot).options }
```
unional: One thing I'm not sure is whether should I guard against `!res.config`

src/core/TypescriptConfig.ts
```diff
@@ -110,8 +110,8 @@ export class TypescriptConfig {
 				const configFileRelPath = this.configFile.replace(this.context.appRoot, "");
 				this.context.log.echoInfo(`Typescript config file:  ${configFileRelPath}`);
 				configFileFound = true;
-				const res = ts.readConfigFile(this.configFile, p => fs.readFileSync(p).toString());
-				config = res.config;
+				const res = ts.readConfigFile(this.configFile, ts.sys.readFile);
+				config = { ...res.config, compilerOptions: ts.parseJsonConfigFileContent(res.config, ts.sys, this.context.appRoot).options };
```
unional: One thing I'm not sure is whether should I guard against `!res.config`

src/core/TypescriptConfig.ts
```diff
@@ -110,8 +110,8 @@ export class TypescriptConfig {
 				const configFileRelPath = this.configFile.replace(this.context.appRoot, "");
 				this.context.log.echoInfo(`Typescript config file:  ${configFileRelPath}`);
 				configFileFound = true;
-				const res = ts.readConfigFile(this.configFile, p => fs.readFileSync(p).toString());
-				config = res.config;
+				const res = ts.readConfigFile(this.configFile, ts.sys.readFile);
```
nchanged: I think you should use fuse detections of tsconfig. It should be there.

unional:
I notice the current change is not sufficient. Will make another change in the morning.

src/core/TypescriptConfig.ts
nchanged: oh maybe you are right. Gimme some time to verify the fix

src/core/TypescriptConfig.ts
```diff
@@ -110,8 +110,8 @@ export class TypescriptConfig {
...
+				config = { ...res.config, compilerOptions: ts.parseJsonConfigFileContent(res.config, ts.sys, this.context.appRoot).options };
 				if (res.error) {
 					this.context.log.echoError(`Errors in ${configFileRelPath}`);
```
unional: Trying to understand how the code works here.

If there are errors, the current code just echo and continue processing.

Is this desirable? I'll need to load the files recursively, so I need to fail the call if one of the files can't be read.

src/core/TypescriptConfig.ts
```diff
@@ -110,8 +110,8 @@ export class TypescriptConfig {
 				const configFileRelPath = this.configFile.replace(this.context.appRoot, "");
 				this.context.log.echoInfo(`Typescript config file:  ${configFileRelPath}`);
 				configFileFound = true;
```
unional: Another concern here.

When there is error, does this still hold true?

unional:
The latest change works.

PR #1412, opened 2018-10-18
upd doc/plugin-api.md

chaimleib:
I added a bunch of extra information and updated all the line numbers in the links to current master. This was a big learning experience, thanks for making FuseBox!

### DIFF
--- a/docs/plugin-api.md
+++ b/docs/plugin-api.md
@@ -1,8 +1,25 @@
 # Plugin API
 
-Plugin API has a very powerful mechanism of manipulating the output.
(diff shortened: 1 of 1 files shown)


docs/plugin-api.md
```diff
@@ -1,8 +1,25 @@
...
+```
+
+Therefore, if your plugin has a few stages of transformations, consider breaking
+it into reusable sub-plugins with standard output formats, and/or even chaining
+in existing plugins to achieve your desired behavior. Like building blocks, the
+possibilities are endless!
+
+## Transforming individual files
+
+Let's take a look a plugin's interface first.
```
nchanged: typo ;-)

PR #1442, opened 2018-11-25
fix: adds stream-browserify as stream polyfill for browser

riglar:
@nchanged - I've made the changes to the stream polyfill (domain is already fixed in #1438 ). I'm currently requiring in `stream-browserify` but let me know if you want this as inline code instead.

I haven't been able to run tests properly as I've ran into issues running `gulp dev` on my machine. Essentially it boils down to this: https://stackoverflow.com/questions/53146394/node-app-fails-to-run-on-mojave-referenceerror-internalbinding-is-not-defined

Ultimately, I would need to upgrade gulp to v4 (and get a new vinyl-fs and new natives).

Guessing this would have breaking changes so I've held off on this. But let me know if you want some PR effort on this.

### DIFF
--- a/modules/stream/index.js
+++ b/modules/stream/index.js
@@ -1,5 +1,5 @@
 if (FuseBox.isServer) {
 	module.exports = global.require("stream");
 } else {
-	module.exports = function() {};
+	module.exports = { Stream: require("stream-browserify") };
 }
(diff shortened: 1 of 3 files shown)


riglar:
Please can this go into release soon?

nchanged:
@riglar sorry, I though I merged it ;-))

Issue #1366, opened 2018-08-28
typescript-plugin-styled-components


PPehrs:
I want to use _typescript-plugin-styled-components_ for styled-components _displayName_ reason

https://www.styled-components.com/docs/tooling#better-debugging is an alternative, but wont work

can someone help me to integrate _typescript-plugin-styled-components_ as plugin or maybe there is another way to _displayName_ for styled-components

nchanged:
https://github.com/Microsoft/typescript-styled-plugin
You mean this one? 

According to docs, this should be good enough
```
{
  "compilerOptions": {
    "plugins": [
      {
        "name": "typescript-styled-plugin"
      }
    ]
  }
}
```

PPehrs:
thanks for this quick answer

it's not the plugin i mean

with the typescript-plugin-styled-components styled-component classNames will be modified from:

<div class="**sc-EHOje** cQxCCl">

to:

<div class="**MyStyledComponent**-EHOje cQxCCl">

we use it already with webpack but fail to use it with fuse-box

https://github.com/Igorbek/typescript-plugin-styled-components

nchanged:
@PPehrs awesome, I think we can port it, since it's only a typescript transformer which FuseBox supports on plugin level.

nchanged:
Is this one still relevant?

PPehrs:
yes, would be great if you can port it, is it possible?

RXminuS:
I put up a bounty if anyone is eager https://www.bountysource.com/issues/62705770-typescript-plugin-styled-components

eds-snippets:
Please close this issue, because it is already solved. Documentation about the plugin you can find [here](https://github.com/fuse-box/fuse-box/blob/master/website/versioned_docs/version-3.6.0/plugins/styled-components-plugin.md). If you have any additional questions about the plugin, just don't hesitate to ask. Thank you.

RXminuS:
Awesome job! You just saved me hours of headaches! @nchanged do I need to do anything for the bounty to pay out?

nchanged:
@RXminuS thank @eds-snippets ;-) You should contact him ;-)

nchanged:
@eds-snippets I guess you should click "claim bounty" on that page

PR #1449, opened 2018-12-06
feat: Add StyledComponentsPlugin #1366

eds-snippets:
I created `StyledComponentsPlugin`. Did two tests as well. Implementation is straight forward, just put the `StyledComponentsPlugin()` into the `getConfig()` plugins section, or with additional options like here:

```
StyledComponentsPlugin({
  getDisplayName: function(filename, bindingName) {
    // typescript-plugin-styled-components gives path to a filename variable
    // we change a name to filename in PascalCase
    filename = filename.replace(/^.*[\\\/]/, '').match(/[a-z]+/gi).map(function (word) {
      return word.charAt(0).toUpperCase() + word.substr(1).toLowerCase()
    }).join('').replace('Tsx', '').replace('Ts', '');
    return filename + '_' + bindingName
  }
}),
```

### DIFF
--- a/src/plugins/StyledComponentsPlugin.ts
+++ b/src/plugins/StyledComponentsPlugin.ts
@@ -13,10 +13,12 @@ export class StyledComponentsPluginClass implements Plugin {
   }
 
   public init(context: WorkFlowContext) {
-    try {
-      require.resolve('typescript-plugin-styled-components');
-    } catch(e) {
-        throw new Error(`Cannot find module 'typescript-plugin-styled-components', install it to devDependency\nnpm install --save-dev typescript-plugin-styled-components`)
+    if (typeof require.resolve === 'function') {
+      try {
+        require.resolve('typescript-plugin-styled-components');
+      } catch(e) {
+          throw new Error(`Cannot find module 'typescript-plugin-styled-components', install it to devDependency\nnpm install --save-dev typescript-plugin-styled-components`)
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,13 +57,16 @@
     "shorthash": "0.0.2",
     "source-map": "^0.7.1",
     "stream-browserify": "^2.0.1",
+    "styled-components": "^4.1.2",
```
nchanged: I don't think we need it in dependencies. It's better to do it the way SassPlugin is done. Require it optionally and throw and error if it's not installed.

eds-snippets: `styled-components` was needed only for testing purposes, I moved it together with `typescript-plugin-styled-components` to `devDependencies`.

eds-snippets:
I added the documentation and removed unnecessary code from the plugin.

src/plugins/StyledComponentsPlugin.ts
```diff
@@ -0,0 +1,28 @@
+import { WorkFlowContext } from "./../core/WorkflowContext";
+import { Plugin } from "../core/WorkflowContext";
+
+const createStyledComponentsTransformer = require('typescript-plugin-styled-components').default;
```
nchanged: Please move that into `init` or some place else inside the plugin, otherwise it will yield an immediate error

Issue #1364, opened 2018-08-28
BabelPlugin with Babel7 support


eddyw:
Now that Babel 7 is out, I was attempting to run a test project with fuse-box and Babel 7. However, [here](https://github.com/fuse-box/fuse-box/blob/master/src/plugins/js-transpilers/BabelPlugin.ts#L101) BabelPlugin explicitly requires `babel-core`. I want to discuss the possibility of allowing the user to define which or from where to include babel core. Kind of like:
```javascript
BabelPlugin({
  babelCore: '@babel/core', // < fallback to `babel-core` if not specified?
  config: {...}
})
```
Since `babel-core` and `@babel/core` have the same API (I think, I took a quick look :P), It would work.
Or, as an alternative, if It can be versioned. Maybe something like:
```javascript
Babel7Plugin({...})
```
Keeping the previous one untouched.

nchanged:
@eddyw sounds legit to me? can you submit a PR? also please update the docs ;-)

eddyw:
@nchanged I have a dirty implementation working on local. It'd take me some time to implement it properly :P

While we're at it, `@babel/preset-env` has a `useBuiltIns` options which can be set to `usage` to include polyfills for things that need to be polyfilled. For instance, with the following:

**Source:**
```javascript
[1, 2, 3].includes(3);
```

**Config:**
```javascript
const client = FuseBox.init({
  cache: false,
  target: "universal@esnext",
  homeDir: "src",
  output: "dist/$name.js",
  useTypescriptCompiler: false,
  tsConfig: "./tsconfig.json",
  sourceMaps: {
    project: true,
    vendor: true,
    inline: true
  },
  ensureTsConfig: true,
  log: false,
  plugins: [
    EnvPlugin({
      NODE_ENV: "development"
    }),
    SourceMapPlainJsPlugin(),
    Babel7Plugin({ // << Just replaces babel-core with @babel/core and `transform` to `transformAsync`
      limit2Project: true,
      config: {
        sourceMaps: true,
        sourceType: "module", // << by default babel will guess
        ast: true, // << this defaults to `false`, I'm not sure if it was like that before Babel 7
        presets: [
          [
            "@babel/preset-env",
            {
              useBuiltIns: "usage",
              targets: "> 0.25%, not dead",
              debug: true,
              modules: "commonjs"
            }
          ]
        ]
      },
      extensions: [".tsx", ".ts", ".js", ".jsx"]
    }),
  ]
})
```
[...]

PR #1453, opened 2018-12-24
Babel7Plugin implementation

eddyw:
Hello there,

I've ported babel plugin to use babel7 core module. The discussion can be found [here](https://github.com/fuse-box/fuse-box/issues/1364)

I've mentioned that I wanted to make babel auto polyfill work together with this plugin. However, to keep this simple, I've only ported the previous plugin to make it work with babel7.

Notes:
- `ast: true` is provided and merged explicitly with the config because on babel7 this property defaults to `false`. To sum up, to keep it consistent with the previous babel plugin.
- Technically there is no need for this plugin to handle loading `.babelrc` since `configFile` or `babelrc` props can be passed as part of the config [(see #configfile)](https://babeljs.io/docs/en/options#configfile) and let babel core handle it. However, this is up for discussion since `babelrc` will look for `.babelrc` relatively to the filename (fuse.js) and not relatively to `homeDir`. It wasn't an issue with the previous babel plugin but on babel7 there is also `babel.config.js` which is a JS file that cannot be loaded as we currently do (in previous plugin and this plugin). However, it can be loaded through `configFile` prop config.

### DIFF
--- a/docs/plugins/babel7.md
+++ b/docs/plugins/babel7.md
@@ -1,0 +1,59 @@
+---
+id: babel7-plugin
+title: Babel7Plugin
+---
+
+## Description
+
+Transpile code to different dialects of javascript.
+
+## Install
+
+This package depends on the `@babel/core` module.
+
+```bash
+yarn add @babel/core @babel/preset-env @babel/preset-react --dev
+npm install @babel/core @babel/preset-env @babel/preset-react --save-dev
+```
+
+##### Parameters
+
+| Name          | Type            | Description                                                                     | ?Default                           |
+| ------------- | --------------- | ------------------------------------------------------------------------------- | ---------------------------------- |
+| config        | `Object`        | when using other fuse-box only properties, babel config is passed in as .config |                                    |
+| limit2project | `boolean`       | to use this plugin across an entire project (including other modules like npm)  | `true`                             |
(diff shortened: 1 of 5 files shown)


src/plugins/js-transpilers/Babel7Plugin.ts
```diff
@@ -0,0 +1,170 @@
...
+		this.handleBabelRc();
+	}
+
+	/**
+	 * @param {File} file
+	 */
+	public transform(file: File) {
+		file.wasTranspiled = true;
+		if (!babel7Core) {
+			babel7Core = require("@babel/core");
```
nchanged: Shall we add exception here? We should say that it's required to install `@babel/core`

docs/plugins/babel7.md
```diff
@@ -0,0 +1,59 @@
+---
+id: babel7-plugin
```
nchanged: Please add this plugin to the list here too.
https://github.com/fuse-box/fuse-box/blob/c8958a7b23178e71b1057cca0555cb70cbe0b67c/website/sidebars.json#L39

nchanged:
Good job! I've added a few comments there.

eddyw:
🙌yay!

nchanged:
@bkniffler `3.7.0-next.13`

Issue #1460, opened 2019-01-19
Window unavailable in worker


bkniffler:
Hey,
this is basically the same issue like webpack had: https://github.com/webpack/webpack/issues/6642
This makes it difficult to use worker with dependencies that rely on window.xxx.

Their solution is an option like this: 
```
output: {
  globalObject: `(typeof self !== 'undefined' ? self : this)`,
},
```

nchanged:
I think use can use Banner plugin for that, but yes we would need a separate plugin for that. It can be based on Banner Plugin

bkniffler:
Good to know. I will let you decide if you want to add that as a quality of life functionality, but I guess the banner plugin will be sufficient, though not super obvious for newcomers. Maybe its enough to have this in the issues, so people who search for that will get on the right track.

nchanged:
@bkniffler well that plugin would make life much easier, since one would need to give it an object. If you want to help and introduce this plugin please go ahead ;-) That would be a real help. You can take a look at BannerPlugin and how it injects custom strings.

bkniffler:
I've created a PR!

bkniffler:
I think the PR is not really required. If you want to make window available in your worker (maybe due to some dependencies you are using check for `typeof window !== "undefined"` to figure out if you are in a browser environment, you could:

```
BannerPlugin(
"var window = typeof self !== 'undefined' ? self : this;"
)
```

PR #1461, opened 2019-01-20
WorkerGlobalScopePlugin

bkniffler:
https://github.com/fuse-box/fuse-box/issues/1460

### DIFF
--- a/docs/plugins/worker-global-scope.md
+++ b/docs/plugins/worker-global-scope.md
@@ -1,0 +1,31 @@
+---
+id: worker-global-scope-plugin
+title: WorkerGlobalScopePlugin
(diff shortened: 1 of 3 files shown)


nchanged:
@bkniffler great, how does it behave in the actual worker? Does it work well?

bkniffler:
Nope, but I just got aware that this is a non issue with fuse. In the end, it is just about `window`, which, logically is not available in a worker. The referenced webpack problem results from hot reloading referencing window automatically in each file, which yields `window is not defined` error.

So, sorry about the fuzz. I could fix the Plugin to make window available, but I guess this should not be necessary in most problems.

Issue #1463, opened 2019-01-21
Babel7 issue with emotion babel plugin


bkniffler:
Hey there,
Here is a minimal example of my issue: https://github.com/bkniffler/fusebox-babel7-issue

Babel7 is now available as a plugin from @eddyw:
https://github.com/fuse-box/fuse-box/pull/1453

There must be a problem with the implementation though. I tried using the emotion babel plugin. While it works great with for example parcel bundler, it will not work with fuse. I included both in the example repository.

Here is the error that appears in console:
```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
    at assertPath (path.js:39:11)
    at Object.parse (path.js:1431:5)
    at getLabel (/Users/bkniffler/Git/fusebox-babel7/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js:177:29)
    at getLabelFromPath (/Users/bkniffler/Git/fusebox-babel7/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js:190:10)
    at transformExpressionWithStyles (/Users/bkniffler/Git/fusebox-babel7/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js:408:19)
    at transformCssCallExpression (/Users/bkniffler/Git/fusebox-babel7/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js:666:31)
    at PluginPass.JSXAttribute (/Users/bkniffler/Git/fusebox-babel7/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js:917:11)
    at newFn (/Users/bkniffler/Git/fusebox-babel7/node_modules/@babel/traverse/lib/visitors.js:193:21)
    at NodePath._call (/Users/bkniffler/Git/fusebox-babel7/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/Users/bkniffler/Git/fusebox-babel7/node_modules/@babel/traverse/lib/path/context.js:40:17)
```

The problematic line is here: https://github.com/bkniffler/fusebox-babel7-issue/blob/master/src/index.tsx#L57

Removing the box around the line will make the problem disappear (the other box above is not an issue).

eddyw:
Hey there,

[...]

PR #1467, opened 2019-01-22
Improve Babel7Plugin and handle JS Babel configuration

eddyw:
Hey,

- solves the issue mentioned at [issues/1463](https://github.com/fuse-box/fuse-box/issues/1463) that happens when `filename` is not passed when transpiling in the configuration. Some plugins wouldn't work before
- improves defaults or required config that must be always set in order to transpile. e.g: `ast: true, code: true, ...`
- validates configuration with `loadOptions` and `loadPartialOptions` from Babel7 core which ensures a valid configuration is provided on init. e.g: presets and/or plugins are resolvable and fails when extra properties or typos in properties' names
- added handling babel JS config files such as `babel.config.js`
- improve lookup for `.babelrc` or `babel.config.js` files relatively to `homeDir`
- added option `fileConfig` which can be `false | string | undefined`. `false` disables lookup for config files, when `string` locates and loads config file
- added more tests
- updated docs

### DIFF
--- a/docs/plugins/babel7.md
+++ b/docs/plugins/babel7.md
@@ -20,14 +20,15 @@ npm install @babel/core @babel/preset-env @babel/preset-react --save-dev
 
 | Name          | Type            | Description                                                                     | ?Default                           |
 | ------------- | --------------- | ------------------------------------------------------------------------------- | ---------------------------------- |
-| config        | `Object`        | when using other fuse-box only properties, babel config is passed in as .config |                                    |
-| limit2project | `boolean`       | to use this plugin across an entire project (including other modules like npm)  | `true`                             |
-| extensions    | `Array<string>` | file extensions to allow with fuse-box                                          | `[".jsx"]`                         |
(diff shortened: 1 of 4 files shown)


bkniffler:
Can this get a quick release again? Would love to try it live!

eddyw:
Merged yay!
@bkniffler I believe it's `3.7.0-next.14` now :)

3cp:
Is there any Babel7Plugin example project?

I have
```
FuseBox.init({
    homeDir: 'src',
    output: 'dist/$name.js',
    plugins: [
      Babel7Plugin({configFile: '.babelrc'}),
    ]
  });
```

But FuseBox still tries to use TypeScript compiler:
```
  FuseBox failed to initialize. Please check that:
      - you have TypeScript installed
          -> if using NPM, run on your terminal:

              npm i -D typescript

          -> if using Yarn, run on your terminal:

              yarn add typescript --dev

      - the TypeScript version installed is >= 3.0
```

Issue #1457, opened 2019-01-04
FuseBox doesn't report diagnostic error on tsconfig.json file


eddyw:
When certain mis-configured options are set on `tsconfig.json` file, those are not actually reported by FuseBox. For instance `{ jsx: "asd" }`

Could we use `ts.convertCompilerOptionsFromJson(...)` instead on `TypescriptConfig.ts`?
This would require to check for `typescript` to be included first.

The reason is because, `File.transpileUsingTypescript` uses `ts.transpileModule` which will transpile code despite errors or mis-configuration in tsconfig file. It does however include a `diagnostics` property (along with `outputText` and `sourceMapText`)

So, we could either check on `TypescriptConfig.ts` or in `File.transpileUsingTypescript` for diagnostic errors on tsconfig file.

Opinions/ideas?
From tsconfig containing `{ ..., jsx: "asd" }`, this is the diagnostic error that should be displayed:
```javascript
Argument for '--jsx' option must be: 'preserve', 'react-native', 'react'.
```

nchanged:
Fair enough , do you know how to fix that? Can you propose a PR?

eddyw:
@nchanged yes, I'm working on it. I'll have a PR in a couple of days or so

A couple of questions:
- we're currently not catching syntax errors. `transpileModule` will try to "repair" broken code and will always transpile but enabling diagnostics we can catch syntax errors. Catching syntax errors is not the same as type errors, so it shouldn't affect transpilation time. It's basically:
```javascript
	private getTranspilationConfig() {
		return Object.assign({}, this.context.tsConfig.getConfig(), {
			fileName: this.info.absPath,
			transformers: this.context.fuse.opts.transformers || {},
			reportDiagnostics: true, // <<< reports syntax and global errors
		});
	}
```
Note:
- global errors are errors without filename such as the ones in the tsconfig file
- syntax errors is just bad JS/TS code that shouldn't be "repaired"
- doesn't do type-checking

Should we do this with a flag? I don't know, such as `reportSyntaxErrors: true` to enable it? or enable always by default?

- Another thing is that, we only check if `typescript` is installed within the `File.ts` on `transpileUsingTypescript` function. However, in `FuseBox.ts` we include `TypescriptConfig.ts` which assumes `typescript` is installed without checking (there is an ts import and the `read` function). How to better handle this? I currently added a `ensureTypescriptInstalled` function to `Utils.ts` which does:
```typescript
[...]

PR #1470, opened 2019-01-22
Report diagnostic errors on compilerOptions - #1457

eddyw:
[issues/1457](https://github.com/fuse-box/fuse-box/issues/1457)

- ensure typescript is installed. Checks on `index.js` since it's the earliest when TS is not used
  - ensures minimal version of TS is `3.0`. Okay? :P
- replace `ScriptTarget` enum with `ts.ScriptTarget`
- parse and validate `compilerOptions`
  - only done once before cached
  - format error|warning|message using TS formatter but displayed as standard FuseBox errors/warnings format
  - once parsed, `compilerOptions` is compatible with `ts.CompilerOptions` which replaces strings with equivalent enums' values. For instance, `jsx: "react"` will be `jsx: ts.JsxEmit.React`. This is helpful for not having mixed configuration (string values | number values)
  - explicitly unset/set `compilerOptions` used in `transpileModule` [transpile.ts#L26](https://github.com/Microsoft/TypeScript/blob/865b3e786277233585e1586edba52bf837b61b71/src/services/transpile.ts#L26) **Edit: removed unset, I realized some of the options could be needed in plugins :P**
    - currently has no effect since `transpileModule` does exactly that. However, it'd spend less time trying to fix and parse internally `compilerOptions`. Probably the time spend is insignificant but I also consider that `transpileModule` could be ported to FuseBox since it's a small function and we could ignore the part where it tries to fix `compilerOptions` internally all over again for each file transpiled.
  - displays warnings|messages|errors but only throws on error to prevents transpilation with broken `compilerOptions`. Rather than having the user figuring out why it doesn't work, the diagnostic messages show a good explanation for every broken property in the config

Notes:
- All current tests pass
- I haven't figured out how to write a test to check when `this.context.fatal` throws with `Error: Invalid 'compilerOptions' settings`, so I checked manually on the playground breaking settings. e.g: `{ target: 'asdasd', jsx: 'asdasd' }`

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -9,6 +9,7 @@ import { Log } from "./Log";
 const userFuseDir = Config.PROJECT_ROOT;
 const stylesheetExtensions = new Set<string>([".css", ".sass", ".scss", ".styl", ".less", ".pcss"]);
 const MBLACKLIST = ["freelist", "sys"];
+export const MINIMAL_TYPESCRIPT_VERSION = "3.0";
 export type Concat = {
 	add(fileName: string | null, content: string | Buffer, sourceMap?: string): void;
 	content: Buffer;
@@ -58,6 +59,40 @@ export function printCurrentVersion() {
 		VERSION_PRINTED = true;
 		const info = getFuseBoxInfo();
 		Log.defer(log => log.echoYellow(`--- FuseBox ${info.version} ---`));
+	}
+}
+
+export function ensureTypescriptInstalled(): never | void {
+	const NO_TYPESCRIPT_MESSAGE = `
+  FuseBox failed to initialize. Please check that:
+      - you have TypeScript installed
+          -> if using NPM, run on your terminal:\n
+              npm i -D typescript\n
+          -> if using Yarn, run on your terminal:\n
+              yarn add typescript --dev\n
+      - the TypeScript version installed is >= ${MINIMAL_TYPESCRIPT_VERSION}
+`;
+	const TYPESCRIPT_NOT_SUPPORTED = `
+  FuseBox failed to initialize. Please check that:
+      - the TypeScript version installed is >= ${MINIMAL_TYPESCRIPT_VERSION}
+`;
+	try {
+		const ts = require("typescript");
+		if (typeof ts.versionMajorMinor === "string") {
+			if (Number(MINIMAL_TYPESCRIPT_VERSION) > Number(ts.versionMajorMinor)) {
+				console.log(TYPESCRIPT_NOT_SUPPORTED);
+				process.exit(1);
+			}
+			return
+		}
+		console.log(NO_TYPESCRIPT_MESSAGE);
+		process.exit(1);
+	} catch (error) {
+		if (error.code === "MODULE_NOT_FOUND") {
+			console.log(NO_TYPESCRIPT_MESSAGE);
+			process.exit(1);
+		}
+		throw error;
 	}
 }
 
--- a/src/core/TypescriptConfig.ts
+++ b/src/core/TypescriptConfig.ts
@@ -93,26 +93,6 @@ export class TypescriptConfig {
 			this.logAllDiagnostics(normalizedCompilerOptions.errors);
 		}
 		this.config.compilerOptions = compilerOptions = normalizedCompilerOptions.options;
-		/**
-		 * Explicitely un-set in `transpileModule`:
-		 * https://github.com/Microsoft/TypeScript/blob/865b3e786277233585e1586edba52bf837b61b71/src/services/transpile.ts#L26
-		 * */
-		compilerOptions.isolatedModules = true;
-		compilerOptions.suppressOutputPathCheck = true;
-		compilerOptions.allowNonTsExtensions = true;
-		compilerOptions.noLib = true;
-		compilerOptions.lib = undefined;
-		compilerOptions.types = undefined;
-		compilerOptions.noEmit = undefined;
-		compilerOptions.noEmitOnError = undefined;
-		compilerOptions.paths = undefined;
-		compilerOptions.rootDirs = undefined;
-		compilerOptions.declaration = undefined;
-		compilerOptions.composite = undefined;
-		compilerOptions.declarationDir = undefined;
-		compilerOptions.out = undefined;
-		compilerOptions.outFile = undefined;
-		compilerOptions.noResolve = true;
 	}
 
 	private verifyTsLib() {
--- a/src/core/TypescriptConfig.ts
+++ b/src/core/TypescriptConfig.ts
@@ -8,13 +8,26 @@ import * as ts from "typescript";
 
 const CACHED: { [path: string]: any } = {};
 
+export interface ICategorizedDiagnostics {
+	errors: ReadonlyArray<ts.Diagnostic & { category: ts.DiagnosticCategory.Error }>,
+	warnings: ReadonlyArray<ts.Diagnostic & { category: ts.DiagnosticCategory.Warning }>,
+	messages: ReadonlyArray<ts.Diagnostic & { category: ts.DiagnosticCategory.Message }>,
+	suggestions: ReadonlyArray<ts.Diagnostic & { category: ts.DiagnosticCategory.Suggestion }>,
+}
+
 export class TypescriptConfig {
 	// the actual typescript config
 	private config: any;
 	private customTsConfig: string;
 	private configFile: string;
-
-	constructor(public context: WorkFlowContext) {}
+	private formatDiagnosticsHost: ts.FormatDiagnosticsHost;
+	constructor(public context: WorkFlowContext) {
+		this.formatDiagnosticsHost = {
+			getCanonicalFileName: file => file,
+			getCurrentDirectory: () => context.homeDir,
+			getNewLine: () => ts.sys.newLine,
+		}
+	}
 
 	public getConfig() {
 		this.read();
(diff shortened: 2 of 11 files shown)


src/tests/QuantumCSSplitting.test.ts
```diff
@@ -229,7 +229,7 @@ export class QuantumCSSSplittingTest {
 				env.fileShouldExist("styles.css.map");
 				env.fileShouldExist("chunks/dcbf7349-ea9fe601.css");
 				env.fileShouldExist("chunks/ea9fe601.css.map");
-				env.fileShouldExist("chunks/c3eff8a7-ea9fe601.js");
```
nchanged: what happened here? ;-)

src/core/TypescriptConfig.ts
```diff
@@ -82,15 +105,98 @@ export class TypescriptConfig {
 			}
 		}
 	}
+
+	public categorizeDiagnostics(diagnostics: ReadonlyArray<ts.Diagnostic> | Readonly<ts.Diagnostic>): ICategorizedDiagnostics {
```
nchanged: how can I test this in action?

eddyw: In playground, modify `tsconfig.json` file with invalid settings such as `{ "target": "asd", "jsx": "asd" }`

src/tests/QuantumCSSplitting.test.ts
eddyw: Hash seems to be different. I changed it back, could you point me in the right direction as where to look?

nchanged: But the tests are failing. Are you running it on windows?

eddyw: Hmm, I realized why. If you are in `master` branch and run `QuantumCSSplitting.test.ts` but before in `File.js` on `transpileUsingTypescript` change it to:
```javascript
				const res = ts.transpileModule(this.contents, this.getTranspilationConfig());
				console.dir(res, { compact: false })
				console.dir(this.getTranspilationConfig())
				return res
```
You'll see this:
```javascript
{
  outputText: '"use strict";\nexports.__esModule = true;\n' +
    'require("./foobar.scss");\n//# ' +
    'sourceMappingURL=foobar.js.map',
  diagnostics: [
    {
      file: undefined,
      start: undefined,
      length: undefined,
      messageText: "Argument for '--target' option must " +
        "be: 'es3', 'es5', 'es6', 'es2015', " +
        "'es2016', 'es2017', 'es2018', " +
        "'esnext'.",
      category: 1,
      code: 6046,
      reportsUnnecessary: undefined
    }
  ],
  sourceMapText: '{"version":3,"file":"foobar.js","sourceRoot":"","sources":["foobar.ts"],"names":[],"mappings":";;AACE,yBAAsB","sourcesContent":["\\n\\t\\timport ' +
    '\\"./foobar.scss\\"\\n"]}'
}
{ compilerOptions:
   { module: 'commonjs',
     target: 'ES7',
     sourceMap: true,
     inlineSources: true,
     jsx: 'react',
     importHelpers: true,
     emitDecoratorMetadata: true,
     experimentalDecorators: true },
  fileName:
   '/Users/huangjx/Desktop/eddyw/Github/fuse-box/.fusebox/virtual-files/1548244391810/foobar.ts',
  transformers: {} }
```
That's because on default, we're using `ES7` .. but it's an invalid target. My branch fixes the issue.
`outputText` in `master` for last test:
```javascript
  outputText: '"use strict";\nexports.__esModule = true;\n' +
    'require("./foobar.scss");\n//# ' +
    'sourceMappingURL=foobar.js.map',
```
`outputText` in my branch:
```  outputText: '"use strict";\nObject.defineProperty(exports, ' +
    '"__esModule", { value: true });\n' +
    'require("./foobar.scss");\n//# ' +
    'sourceMappingURL=foobar.js.map',
```

eddyw: I reverted last commit to which changes the hash on Quantum test, we were not transpiling to the correct target.
FuseBox default `compilerOptions` uses `{ target: "ES7" }` which should have been `{ target: "es2016" }`. This branch uses `ts.ScriptTarget` and parses and validates tsConfig, so compilerOptions passed are `{ target: 3 }` which corresponds to `es2016`|`es7`
[File.ts#L18](https://github.com/fuse-box/fuse-box/blob/master/src/core/File.ts#L18)

Since errors were not checked before in CompilerOptions, those were not caught. And since `transpileModule` always transpiles regardless of errors, the output now it wrong:
```javascript
exports.__esModule = true;
require("./foobar.scss");
```
Expected:
```javascript
Object.defineProperty(exports, "__esModule", { value: true });
require("./foobar.scss");```

Comment (eddyw) [reply src/tests/QuantumCSSplitting.test.ts]: Local simple test:
```const ts = require('ts');
const a = ts.transpileModule('export default 123', { compilerOptions: { module: 'commonjs', target: 'es2016' } }) // <<< expected
const b = ts.transpileModule('export default 123', { compilerOptions: { module: 'commonjs', target: 'ES7' } }) // <<< fails```

Comment (eddyw) [reply src/tests/QuantumCSSplitting.test.ts]: @nchanged ping 🙈

Comment (nchanged) [reply src/tests/QuantumCSSplitting.test.ts]: @eddyw sorry ;-) thanks for pinging. Will check it out today

Comment (eddyw) [reply src/tests/QuantumCSSplitting.test.ts]: @nchanged ping 🙈
Anyway to help?

File: src/Utils.ts [hunk 3]
```diff
@@ -61,6 +62,40 @@ export function printCurrentVersion() {
...
+			return
+		}
+		console.log(NO_TYPESCRIPT_MESSAGE);
+		process.exit(1);
+	} catch (error) {
+		if (error.code === "MODULE_NOT_FOUND") {
+			console.log(NO_TYPESCRIPT_MESSAGE);
+			process.exit(1);
+		}
+		throw error;
```
nchanged: Do you think we can make it more beautiful ?
screenshot 2019-01-28 20 56 49

I think we can just take the first line from the stack trace and print it nicely. like we print that red message

nchanged:
@eddyw some small nitpick, but other than that it works flawlessly! Good job!

eddyw:
@nchanged 
- Prettier error messages
  - suggests where to look if tsconfig file was used/found or if set over-writes in `tsConfig` fuseBox option

image

- docs suggest that `tsConfig` can be an array of compilerOptions, however it was typed only as `string`, I made some small changes to allow multiple arrays on `tsConfig` and typed the array for compilerOptions, so the user will know which options they can set in `tsConfig: [{ [CompilerOptions] },...]`
- pulled from master and merged into my branch
  - fixed conflicts, mostly in `ScriptTarget` uses

nchanged:
@eddyw https://github.com/fuse-box/fuse-box/commit/6c6c562b9abb37cb6eff9472c2df06a636898d1c

It seems that we all tried to fix the same issue here ;-)

nchanged:
@eddyw I can't really merge it, since there will be errors when a `tsconfig.json` file is generated. Could you drop the recent changes with targets, and we can see how we can improve ScriptTarget in a different PR?

eddyw:
@nchanged commented at the same time above haha
Yes, I noticed the change. However, in my branch I'm relying on `ts.ScriptTarget` which is more independent on TS version installed by the user. e.g: if we happen to have `ES2019` later, it'd still work working change and the check on compiler options would suggest all available targets.

eddyw:
@nchanged commented at the same time again :P
If I drop changes with targets, the errors may displayed about the `target` may not be accurate (according to TS version installed by user)

Maybe I could help improve when the `tsconfig` file is generated?

nchanged:
@eddyw yeah we do comment at the same time ;-) Of course, please read carefully my commit over there, (btw, there is console.log which needs to be removed) Make sure that an automatically generated `tsconfig.json` has the correct target (default is es2018 - I think it'll do just fine)

And squash yours commits + rebase. Github tells me that I cannot rebase.

src/tests/QuantumCSSplitting.test.ts
nchanged: There is some very annoying timing issue (racing condition) with this test which bugs me from time to time (

nchanged:
@eddyw 
how about this condition (and in a few more places)

https://github.com/fuse-box/fuse-box/blob/6c6c562b9abb37cb6eff9472c2df06a636898d1c/src/core/File.ts#L207-L211

eddyw:
> @eddyw
> how about this condition (and in a few more places)
> 
> [fuse-box/src/core/File.ts](https://github.com/fuse-box/fuse-box/blob/6c6c562b9abb37cb6eff9472c2df06a636898d1c/src/core/File.ts#L207-L211)
> 
> Lines 207 to 211 in [6c6c562](/fuse-box/fuse-box/commit/6c6c562b9abb37cb6eff9472c2df06a636898d1c)
> 
>  public setLanguageLevel(level: ScriptTarget) { 
>  	if (this.languageLevel < level) { 
>  		this.languageLevel = level; 
>  	} 
>  }

It works the same way as when there was a custom `ScriptTarget`

eddyw:
@nchanged 
there is this function `normalizeAndValidateCompilerOptions` which is called to normalize compilerOptions, so if `target` is string (e.g: 'es5' or 'ES5') after analyze and normalize, we'd have `target: 1` (or equivalent enum numeric value always). Before FuseBox showed, for example `Typescript script target: 1` but I also changed that to display `ts.ScriptTarget[config.compilerOptions.target]` instead since `target` will be numeric always, we get the string value from enum, in the previous case `ES5`
https://github.com/fuse-box/fuse-box/blob/1f103c400199d19dc405a975ac252902c0ed5344/src/core/TypescriptConfig.ts#L89-L97
Btw, squashed and rebased some commits

nchanged:
@eddyw still can't rebase it ;-(

nchanged:
@eddyw there is a problem somewhere. FuseBox transpiles node modules using the latest preset ignoring the config.. (

PR #1472, opened 2019-01-24
feat($devServer): Use fusebox logger for HTTPServer logging

TylorS:
I have a use case where I'd like for the `Development Server Running...` log message to be disabled while other information populates the user's terminal, and I found it a bit surprising it wasn't disabled when FuseBox is created with `log: false`.

Additionally, thanks for working on such an awesome project 🌹. I greatly appreciate the strong TypeScript support as it aids me in my own TypeScript tooling projects. I plan to contribute more in the future as I can, so please let me know if there is anything I can do to improve this PR.


nchanged:
@TylorS please do so ;-) Any kind of help is appreciated!

Issue #1474, opened 2019-01-24
FuseBox doesn't resolve "autoImport" if target is server


eddyw:
Reading the docs, there is no mention of `autoImport` not .. auto importing when `target` is `server`. I'm not sure if it was intentional or docs need some clarification.

I was testing a project with `target` set to `browser` which does:
```javascript
autoImport: {
  Promise: 'bluebird'
}
```
However, when setting target to `server`, my script fails on runtime.

nchanged:
Did you clear the cache? That's the most common problem. Please share you config too

eddyw:
with `cache: false`:
Config:
```js
const fuse = FuseBox.init({
  homeDir: '.',
  output: 'dist/$name.js',
  target: 'server',
  sourceMaps: { project: true, vendor: false },
  useTypescriptCompiler: true,
  ensureTsConfig: false,
  cache: false,
  allowSyntheticDefaultImports: true,
  log: {
    showBundledFiles: false,
  },
  autoImport: {
    Promise: 'bluebird',
  },
  plugins: [
    //new AutoImport({
    //  Promise: 'bluebird',
    //}),
    Babel7Plugin({
      config: {
        sourceMaps: true,
        presets: [
          [
            '@babel/preset-env',
            {
              debug: true,
              modules: 'commonjs',
              shippedProposals: true,
              targets: { node: 'current' },
              useBuiltIns: 'usage',
              exclude: ['es6.promise'],
            },
          ],
        ],
      },
    }),
  ],
})
```
I believe this are the lines responsible:
https://github.com/fuse-box/fuse-box/blob/489fb29287b012e95b96c99d3aae7cac0f786ac1/src/core/WorkflowContext.ts#L423-L432
In FuseBox constructor, it calls:
```js
this.context.initAutoImportConfig(opts.natives, opts.autoImport);
```
Maybe the nested if block was supposed to be after `if (this.target !== "server")` block?

nchanged:
@eddyw yeah..indeed. The problem is that some of the internal modules need to be auto imported when the target is browser... If user specifies his own custom config, we need to feed users' config and ignore everything else. Do you think you can try to fix it?

PR #1475, opened 2019-01-25
Fix autoImport when target is server

eddyw:
Fixes [issues/1474](https://github.com/fuse-box/fuse-box/issues/1474)


nchanged:
@eddyw there is some test failing badly? Could you check it?

eddyw:
@nchanged is it possible to re-run the tests of CI?
All my tests are passing locally `npm test`. Sometimes one or two fail with timeout, re-running again they pass :/

PR #1488, opened 2019-02-07
FuseBox 4.0 Pending update

nchanged:
Pending update to make sure 3.7 is stable enough and does not require any patches.


.editorconfig
```diff
@@ -4,15 +4,11 @@ root = true
...
+max_line_length=120
 trim_trailing_whitespace = true
-
 [*.md]
-trim_trailing_whitespace = false
-
-[*.ts]
-indent_style = tab
-max_line_length = 120
+trim_trailing_whitespace = falses
```
firede: Reminding, here is a typo error: `falses` -> `false`

nchanged: thanks!

Issue #1480, opened 2019-01-31
Discussion: FuseBox TypeScript type-checking


eddyw:
Hey,

I've been working on integrating TS type-check on FuseBox since it's something all known bundlers seem to be missing out of the box. Since FuseBox has first class support for TS, it just seemed right. So these are the ideas I had, broken in several steps:
- Global type-check of `tsconfig` file (merged [issues/1457](https://github.com/fuse-box/fuse-box/issues/1457))
- Improve the check of `tsconfig` file (merged [pull/1484](https://github.com/fuse-box/fuse-box/pull/1484)):
  - use `ts.readConfigFile` to read the `tsconfig.json` file which reports errors in JSON file using diagnostics which will help us map and display nicely which line is wrong (like missing `,`). Since previous was merged, this is next :)
  - I realized that some diagnostics should be ignored, such as errors in `rootDir` and `files` which FuseBox doesn't use or need, so reporting those errors is irrelevant
  - `transpileModule` is now reliable since now we're checking `compilerOptions`. However, it doesn't report syntax errors, `Acorn` parser does.. at some extent but doesn't catch always. I don't know if it'd affect performance implementing it, so I left the task to `Acorn` but probably it should be enabled if full-type check
- TS type-checking - I'm working on a plugin but as @nchanged mentioned in [here](https://github.com/fuse-box/fuse-box/issues/1479#issuecomment-458831563) if it works well and it's fast it could be integrated behind a flag. This is currently a blocker until merged -> [hmt-request-missing-deps](https://github.com/fuse-box/fuse-box/commit/a318900677456a09eb95eb85810ce4f377952b18) since resolved dependencies is needed
- resolve tsconfig `paths` (WIP [pull/1495](https://github.com/fuse-box/fuse-box/pull/1495))

I've been doing some research on how to fasten type-checking using TS language services. This is what I took:
- There is a document registry which keeps source files versioned and re-uses the files that have not changed. Ideally, we should cache all files `{ source: [TS original code], version: [number], diagnostics: [previous errors/warnings] }` and re-hydrate the registry from cache, so it'd spend less time looking for files and analysing
[...]

PR #1495, opened 2019-02-11
Implementation of tsconfig paths resolution

eddyw:
[issues/1480](https://github.com/fuse-box/fuse-box/issues/1480)

### DIFF
--- a/src/core/WorkflowContext.ts
+++ b/src/core/WorkflowContext.ts
@@ -75,10 +75,6 @@ export class WorkFlowContext {
 
 	public dynamicImportsEnabled = true;
 
-	public automaticAlias = true;
-
-	// public automaticTSPathAlias = false;
-
 	public tsPathsRegExp?: RegExp;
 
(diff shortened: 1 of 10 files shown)


src/core/WorkflowContext.ts
```diff
@@ -272,9 +281,50 @@ export class WorkFlowContext {
...
+		}
+
+		return { requireStatement, replaced };
+	}
+
 	public replaceAliases(requireStatement: string): { requireStatement: string; replaced: boolean } {
 		const aliasCollection = this.aliasCollection;
 		let replaced = false;
+
+		if (this.automaticTSPathAlias && this.tsPathsRegExp && this.tsPathsRegExp.test(requireStatement)) {
```
nchanged: Shall we also check also that we are in a typescript file?  We don't want to do that on node_modules

eddyw: Added that :)

PR #1496, opened 2019-02-11
feat: SassPlugin supports both, node-sass and sass (with fibers optional)

sod:
dart sass (https://github.com/sass/dart-sass) is now stable and has a javascript version published that can be installed via `npm install sass`. Dart sass it the official implementation of sass and new features will be there first. The api is the same, so you can drop in either of both.

It has an optional dependency `fibers` that increases performance.

node-sass is stil faster. But as your fuse-box caching makes transpiling a rare occurance anyway, it would be nice if the SassPlugin could use both, `node-sass` and `sass`, depending on what the user installed.

@nchanged

### DIFF
--- a/docs/plugins/sass.md
+++ b/docs/plugins/sass.md
@@ -15,6 +15,16 @@ yarn add node-sass --dev
 npm install node-sass --save-dev
 ```
 
+Note: Instead of `node-sass`, you may also use the reference implementation 
+[`sass`](https://github.com/sass/dart-sass) together with its optional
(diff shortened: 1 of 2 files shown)


nchanged:
Brilliant! Could you change the destination to branch `v4`

sod:
rebased to v4 and added mention in sass docs

nchanged:
@sod could you rebase again please? So many commits there.

sod:
done

PR #1501, opened 2019-02-18
chore: update fs-extra dependency

jasonlangdon-sky:
This update addresses https://github.com/jprichardson/node-fs-extra/commit/ddc1a2f9d42da77342d58f6fd86942a231f5f27f which was introduced into fuse-box when fs-extra was updated in this PR https://github.com/fuse-box/fuse-box/pull/1341/files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,36 @@
+{
+  "name": "fuse-box",
+  "version": "3.7.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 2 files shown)


nchanged:
@jasonlangdon-sky great! could you make a PR against `v4` branch?

jasonlangdon-sky:
Will there be no more 3.x releases?

nchanged:
@jasonlangdon-sky no, only `4.0` which you can install already now `fuse-box@next`

PR #1531, opened 2019-05-24
fix: Synthetic default export polyfill does not add __esModule property

Gustorn:
This should fix the issue with the Babel import polyfill (and any similar helper functions)

### DIFF
--- a/modules/fuse-loader/index.js
+++ b/modules/fuse-loader/index.js
@@ -303,13 +303,13 @@ function $trigger(name, args) {
 }
 // NOTE: Should match syntheticDefaultExportPolyfill in fuse-box-responsive-api/index.js
 function syntheticDefaultExportPolyfill(input) {
-  if (input === null || ['function', 'object', 'array'].indexOf(typeof input) === -1) {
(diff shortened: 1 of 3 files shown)


modules/fuse-loader/index.js
```diff
@@ -303,24 +303,30 @@ function $trigger(name, args) {
 }
 // NOTE: Should match syntheticDefaultExportPolyfill in fuse-box-responsive-api/index.js
 function syntheticDefaultExportPolyfill(input) {
-  if (
-    input === null ||
-    ['function', 'object', 'array'].indexOf(typeof input) === -1 ||
-    (input && Object.prototype.hasOwnProperty.call(input, 'default')) // use hasOwnProperty to avoid triggering usage warnings from libraries like mobx
```
nchanged: What happened to this line and the warning about mobx?

Gustorn: Sorry about the missing comments, deleting those was an accident, I meant to re-add them once the fix was complete. I changed the checks back to use `Object.prototype.hasOwnProperty` but removed the `input &&` part since it is unnecessary as we return from the function if the input is either `null` or `undefined`.

PR #1565, opened 2019-07-12
typedoc implementation

paxperscientiam:
Hello, this is rough. The docs seemed to be built, but I have no idea how to audit them. 

I am unsure how to contend with this error: `Error: Cannot write file '/Users/ramos/repos/fuse-box/src/main/__tests__/cases/projects/src1/index.js' because it would be overwritten by multiple input files.`


nchanged:
that's cool but we need to select the files we want to document, there is too much noise.

paxperscientiam:
OK, I'll play around.

paxperscientiam:
@nchanged , starting to get hang of it. I was able to hide all private properties and target specific files.

Are we moving in the right direction?

paxperscientiam:
Nuking because I botched something. I'll make a new PR.

PR #1587, opened 2019-08-22
feat(CustomTransformers): Implemented support for user-provided TS transformers

kyr0:
As discussed on gitter, this PR implements the changes required to freely configure user-provided custom `before`, `after`, `afterDeclaration` stage TypeScript compiler transformation functions. 

The implementation respects the differences of the development and production pipeline and comes with a playground-test to end2end-test it's functionality. 

Core-provided transformers take precedence and so their transformations can be altered by user-provided transformers if needed. This gives ultimate freedom for TypeScript-based projects.

Bildschirmfoto 2019-08-22 um 03 22 34

Bildschirmfoto 2019-08-22 um 03 22 20

### DIFF
--- a/playground/custom_transformer/fuse.ts
+++ b/playground/custom_transformer/fuse.ts
@@ -1,0 +1,97 @@
+import { fusebox } from '../../src/core/fusebox';
+import * as ts from 'typescript';
+
+const isProd = process.argv.indexOf('--prod') > -1;
+
+function transformNode(node) {
+
+  // @ts-ignore
+  if (ts.isPropertyAccessExpression(node) && node.expression) {
+    // @ts-ignore
+    if (node.expression.expression && node.expression.expression.getText() && node.expression.expression.getText() === 'process' &&
+      // @ts-ignore
+      node.expression.name && node.expression.name.getText() && node.expression.name.getText() === 'env') {
+
+      // @ts-ignore
+      const envVarName = node.name.getText();
+      const envVarValue = process.env[envVarName];
+
+      return ts.createStringLiteral(envVarValue || '');
+    }
+  }
+}
+
+function exampleProdBeforeTransformer<T extends ts.Node>(): ts.TransformerFactory<T> {
+  return (context) => {
+
+    console.log();
+    console.log('===> custom prod mode, before stage transformer called');
+    console.log();
+
+    const visit: ts.Visitor = (node) => {
+
+      const transformedNode = transformNode(node);
+
+      if (transformedNode) {
+        return transformedNode;
+      }
+      return ts.visitEachChild(node, (child) => visit(child), context);
+    };
+    return (node) => ts.visitNode(node, visit);
+  }
+}
+
+function exampleDevAfterTransformer<T extends ts.Node>(): ts.TransformerFactory<T> {
+  return (context) => {
+
+    console.log();
+    console.log('===> custom dev mode, after stage transformer called');
+    console.log();
+
+    const visit: ts.Visitor = (node) => {
+
+      const transformedNode = transformNode(node);
+
+      if (transformedNode) {
+        return transformedNode;
+      }
+      return ts.visitEachChild(node, (child) => visit(child), context);
+    };
+    return (node) => ts.visitNode(node, visit);
+  }
+}
+
+const fuse = fusebox({
+  target: 'browser',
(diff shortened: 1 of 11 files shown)


src/production/stages/transpileStage.ts
```diff
@@ -95,7 +95,22 @@ function transpileProductionModule(props: ITranspileStageProps, prodModule: Prod
   const result = ts.transpileModule(text, {
     fileName: prodModule.module.props.absPath,
     compilerOptions: compilerOptions,
-    transformers: { after: [moduleTransformer(props, prodModule)] },
+
+    transformers: {
+
+      // merge in any custom transformers (user-provided)
+      ...ctx.customTransformers,
```
nchanged: so you can't add `after` transform for production?

nchanged: sorry, I see you applied it manually below ;-)

src/transform/tsTransform.ts
```diff
@@ -75,12 +73,28 @@ export function tsTransform(props: ITypescriptTransformProps): ts.TranspileOutpu
...
   }
+
   return ts.transpileModule(props.input, {
     fileName: props.fileName,
     compilerOptions: props.compilerOptions,
-    transformers: { after: after },
+    transformers: {
+
+      // merge in any custom transformers (user-provided)
+      ...props.transformers,
```
nchanged: I think that spread operator is very expensive to be applied each time on every single file. Unless you convince me there is no difference performance wise ;-)

src/config/IPublicConfig.ts
```diff
@@ -51,6 +53,7 @@ export interface IPublicConfig {
   stylesheet?: IStyleSheetProps;
   cache?: boolean | ICacheProps;
   tsConfig?: string | IRawCompilerOptions;
+  customTransformers?: CustomTransformers;
```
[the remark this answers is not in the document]
kyr0: Because the privateConfig takes it's config values from the publicConfig and so it's propagated to the production pipeline: https://github.com/kyr0/fuse-box/commit/cfabfef38a0af9e2ea768e742ee75e7436f1f1b8#diff-68ccf37364cb8ccba6bdc3fa98aaba84R39

And for sure, to be able to set it from the outside:
https://github.com/kyr0/fuse-box/commit/cfabfef38a0af9e2ea768e742ee75e7436f1f1b8#diff-8ec425bf0945566e64ad90d79c388262R75

playground/custom_transformer/fuse.ts
```diff
@@ -0,0 +1,97 @@
...
+  entry: 'src/index.ts',
+  modules: ['modules'],
+  logging: {
+    level: 'succinct',
+  },
+  webIndex: {
+    template: 'src/index.html',
+  },
+  sourceMap: true,
+  customTransformers: isProd ? {
```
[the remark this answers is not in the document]
kyr0: Sure, await my PR update :) btw. transformers in the config might clutter it, maybe I should refactor them out in separate files.

src/transform/tsTransform.ts
kyr0: Yeah you're right, it can be up to 3x slower than Array.concat for the case of the Array spread. But according to https://thecodebarbarian.com/object-assign-vs-object-spread.html Object spread operator is faster than Object.assign. I have no evidence on that tho. 

But, tbh. this is loaded only once and not executed every time/at bundle time. How fast do you think, Array.concat or the object merge would be on this? I guess all the operations are waaay below 1 ms in combination. 

Also, for an alternative impl., we would need a deep merge here to have the same outcome. For a good deep merge, we'd need loops or, even worse, external libs like lodash. TypeScript can generate these loops for us by using the ... spread on the object literal. I would always go for high performance for every operation inside of a build step, but in this specific case I would rather go for readability, less and more elegant code. What do you think?

nchanged:
Any updates? It's okay if you don't have time to do this :) we can close it

kyr0:
Yes, there is a new [PR](https://github.com/fuse-box/fuse-box/pull/1601) #1601  with all the changes that we discussed. I was busy getting well again after being sick for some days and afterwards I was busy working on the [general purpose HMR polyfill](https://github.com/vegarringdal/custom-elements-hmr-polyfill) for web component together with @vegarringdal. Closing this now.

PR #1590, opened 2019-08-23
 fix: configParser - stop base config from overriding main

vegarringdal:
Here we go again.
Added fixes, playground and tests.
Tests on configParser was successfull

@nchanged 
Please have new look at the new PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ _playground
 .release
 .settings
 .tmp
-.vscode
+
 build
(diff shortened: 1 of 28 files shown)


vegarringdal:
I need to add little more... Ill tell you when Im done

vegarringdal:
k, new playground tested and vscode launch with vscode debugger works.
Now I it shoudl be ready

PR #1601, opened 2019-08-31
V4 custom transformers plugin

kyr0:
A plugin for custom TypeScript AST transformation


src/transform/tsTransform.ts
```diff
@@ -1,86 +1,31 @@
...
-    after.push(moduleTransformer());
+  if (!props.transformers) {
+    props.transformers = {};
   }
+
   return ts.transpileModule(props.input, {
     fileName: props.fileName,
     compilerOptions: props.compilerOptions,
-    transformers: { after: after },
+    transformers: {
```
nchanged: Do you think it makes sense to create a function `tsTranspilerModule` and pass properties. This way we could also test it. There are 2 places where we call `ts.transpileModule`

src/production/stages/transpileStage.ts
```diff
@@ -95,10 +33,16 @@ function transpileProductionModule(props: ITranspileStageProps, prodModule: Prod
     compilerOptions.inlineSources = true;
   }
 
+  const customTransformers = ctx.customTransformers || {};
+
   const result = ts.transpileModule(text, {
```
nchanged: Also here. ;-) We could could pass `customTransformers` and test it

PR #1603, opened 2019-09-01
feat(tsTransformModule): Added tests

kyr0:


src/transform/__tests__/tsTransformModule.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    return node => ts.visitNode(node, visit);
+  };
+
+  beforeAll(() => {
+    const config = initTypescriptConfig({});
+    compilerOptions = config.compilerOptions;
+  });
+
+  it('should transform module without any custom transformers', () => {
+    const transformedModule = tsTransformModule(
```
nchanged: Do you think it makes sense to have one props like everywhere else? It will be easier if need to refactor it. Also pass customTransformers as is ;_)

kyr0: Sry I don't get what you mean :) I like the expressiveness of the tests instead of looping over configurations if you mean that. But it's correct that it's easier to change. What do you mean by pass customTransformers as is?

nchanged:
```
tsTransformModule(props : ITsTransformModule) {}
```

like everywhere in the code, a function accepts only one argument. That's not about the test per se, just about the function itself ;-)

kyr0: Hmm, that changes quite alot. Why did you merge the initial PR in? ;)

kyr0: Then we'd basically change the interface, both calls, all tests  create the interface. Should we do that all in this PR?

nchanged: Ok, let's merge this one, and change after that ;-) I am ok with that. It's just better if we keep it consistent everywhere =)

PR #1611, opened 2019-09-03
fix: add try/catch for postCss CssSyntaxErrors

vegarringdal:
Added try/catch

If we had invalid css fusebox shut down an quit, se last S after `green;`
image

Error:
image

Added try/catch and logger message 😄 
Now it will not quit and user can fix error
image


vegarringdal:
Great, its a start, most important part is that fusebox keeps on rolling if errors happen :-)

PR #1615, opened 2019-09-05
No tsconfig warning

kyr0:


src/fuse-log/fuseLog.ts
```diff
@@ -48,50 +44,27 @@ export class FuseLog {
   }
 
   warn(...args: any) {
```
nchanged: forgot to change it?

src/fuse-log/fuseLog.ts
```diff
@@ -48,50 +44,27 @@ export class FuseLog {
...
-      str += `<yellow>${message}</yellow>`;
-    } else {
-      str += `<bold>@warning <yellow>${message}</yellow></bold> `;
-    }
+    const { message, vars } = parseArguments(args);
+    const str = `${this.indent}<bold>@warning <yellow>${message}</yellow></bold> `;
     this.log('warn', codeLog(str, vars));
   }
 
   success(...args: any) {
```
nchanged: same here?

PR #1619, opened 2019-09-06
Awaitable ICT

kyr0:
This PR adds the functionality to async/await for callbacks registered with an interceptor. 

Based on this implementation, fuse-box can decide internally where it wants to wait for all registered callbacks to finish before it continues it's work.

The new API `awaitForSync` is backward compatible with all callbacks registered using `on` and `awaitOn`.
The old API `sync` can handle callbacks registered by `awaitOn` in theory, but it is likely that the fuse-box internal code will fail because it doesn't expect a Promise as a return value. You really have to use the API wrong to come into this situation. But when we add a list of events to the documentation, we should also document which ones shall be sync and async

### DIFF
--- a/src/interceptor/interceptor.ts
+++ b/src/interceptor/interceptor.ts
@@ -4,8 +4,10 @@ interface TypedInterceptor<T> {
   getPromises: () => Array<any>;
   promise: (fn: () => Promise<any>) => void;
   resolve: () => Promise<any>;
-  on<K extends keyof T>(s: K, props: (props: T[K]) => T[K]);
-  sync<K extends keyof T>(s: K, props: T[K]): T[K];
+  on<K extends keyof T>(key: K, fn: (props: T[K]) => T[K]);
+  sync<K extends keyof T>(key: K, props: T[K]): T[K];
+  awaitForSync<K extends keyof T>(key: K, props: T[K]): Promise<T[K]>;
+  awaitOn<K extends keyof T>(key: K, fn: (props: T[K]) => Promise<T[K]>);
 }
 export type MainInterceptor = TypedInterceptor<InterceptorEvents>;
 export function createInterceptor(): MainInterceptor {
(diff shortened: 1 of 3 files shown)


src/interceptor/interceptor.ts
```diff
@@ -4,8 +4,10 @@ interface TypedInterceptor<T> {
   getPromises: () => Array<any>;
   promise: (fn: () => Promise<any>) => void;
   resolve: () => Promise<any>;
-  on<K extends keyof T>(s: K, props: (props: T[K]) => T[K]);
-  sync<K extends keyof T>(s: K, props: T[K]): T[K];
+  on<K extends keyof T>(key: K, fn: (props: T[K]) => T[K]);
+  sync<K extends keyof T>(key: K, props: T[K]): T[K];
+  awaitForSync<K extends keyof T>(key: K, props: T[K]): Promise<T[K]>;
```
nchanged: i am having hard time understanding the terminology here ;-)
`awaitForSync` but you are you using an async function. `waitFor`  

And both functions return a promise, that's a bit misleading.

kyr0: awaitForSync must return a Promise, otherwise we can't await the end result. And each callback must return a Promise, otherwise we can't awair their single result (in awaitForSync). Basically, I would have said that we could also just change sync and on to always do their work Promise based and the callbacks to always be async functions. Then we get rid of all problems and have a concise API

src/interceptor/interceptor.ts
```diff
@@ -46,5 +51,22 @@ export function createInterceptor(): MainInterceptor {
       }
       return props;
     },
+    awaitOn: on,
+    awaitForSync: async function(key: string, props: any) {
+      if (subscriptions.has(key)) {
+        const fns = subscriptions.get(key);
+        const responses = [];
+
+        for (let i = 0; i < fns.length; i++) {
```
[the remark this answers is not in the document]
kyr0: I'll change that

kyr0:
Implemented the name changes suggested and also the for loop change. For the `send` instead of `sendAwait` the idea was: `await sendAwait(...)` would probably be duplicated emphasizing :)

src/interceptor/interceptor.ts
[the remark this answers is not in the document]
kyr0: Done

src/interceptor/interceptor.ts
kyr0: Done

PR #1620, opened 2019-09-06
WebIndex: Async/await ICT support

kyr0:
Allows for async WebIndex plugins. Added documentation for `webIndex`. 

This PR relies on #1619 to be merged into master before.

### DIFF
--- a/src/web-index/webIndex.ts
+++ b/src/web-index/webIndex.ts
@@ -72,6 +72,7 @@ export function getEssentialWebIndexParams(config: IWebIndexConfig | boolean, lo
     distFileName,
     publicPath,
     templateContent,
+    templatePath,
   };
 }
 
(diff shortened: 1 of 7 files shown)


src/web-index/webIndex.ts
```diff
@@ -97,12 +97,23 @@ export function createWebIndex(ctx: Context): IWebIndexInterface {
...
         css: cssTags.join('\n'),
       };
-      let contents = replaceWebIndexStrings(readFile(opts.templatePath), scriptOpts);
+
+      let fileContents = readFile(opts.templatePath);
+
+      const pluginResponse = await ctx.ict.awaitForSync('before_webindex_write', {
+        filePath: opts.templatePath,
+        fileContents,
+        bundles,
```
nchanged: I think we need to add script tags and cssTags here too (or just a list of css paths and js paths), don't you think?

kyr0: I'll look into it, but aren't these part of the buldes?

kyr0: Added the missing scriptTags and cssTags below, also updated docs

PR #1621, opened 2019-09-07
pluginConsolidate

kyr0:
This PR is based on PR #1619 and PR #1620 in this particular order :)

### DIFF
--- a/src/plugins/core/plugin_consolidate.ts
+++ b/src/plugins/core/plugin_consolidate.ts
@@ -4,11 +4,11 @@ import * as consolidate from 'consolidate';
 
 export function pluginConsolidate(engine: string, options: any) {
   return async (ctx: Context) => {
-    ctx.ict.awaitOn(
+    ctx.ict.waitFor(
       'before_webindex_write',
       async (props: { filePath: string; fileContents: string; bundles: Array<IBundleWriteResponse> }) => {
         if (typeof consolidate[engine] !== 'function') {
-          ctx.log.error(`The template engine you selected is not available in consolidate: ${engine}`);
+          ctx.fatal(`The template engine you selected is not available in consolidate: ${engine}`);
         }
 
         try {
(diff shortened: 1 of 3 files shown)


src/plugins/core/plugin_consolidate.ts
```diff
@@ -0,0 +1,30 @@
...
+            ...options,
+            bundles: props.bundles,
+          });
+          props.fileContents = processedTemplate;
+        } catch (e) {
+          ctx.log.error(`Error processing the web-index template using consolidate: ${e.message}`);
+        }
+
+        return {
+          ...props,
```
nchanged: why spread? xD `return props`

src/plugins/core/plugin_consolidate.ts
```diff
@@ -0,0 +1,30 @@
...
+import { IBundleWriteResponse } from '../../bundle/Bundle';
+import * as consolidate from 'consolidate';
+
+export function pluginConsolidate(engine: string, options: any) {
+  return async (ctx: Context) => {
+    ctx.ict.awaitOn(
+      'before_webindex_write',
+      async (props: { filePath: string; fileContents: string; bundles: Array<IBundleWriteResponse> }) => {
+        if (typeof consolidate[engine] !== 'function') {
+          ctx.log.error(`The template engine you selected is not available in consolidate: ${engine}`);
```
nchanged: how about ctx.fatal? I would make sense to stop the execution completely.

src/plugins/core/plugin_consolidate.ts
kyr0: To clone it xD Because we can! 😂

src/plugins/core/plugin_consolidate.ts
kyr0: I agree 👍

src/plugins/core/plugin_consolidate.ts
nchanged: Nah don't clone them in fact I made a mistake in the first place, there is no need in returning anything

kyr0: Sry if it wasn't clear, but that was a joke ;) The clone was just something that remained from an old impl where I merged several objects

src/plugins/core/plugin_consolidate.ts
kyr0: Done

src/plugins/core/plugin_consolidate.ts
kyr0: Done

Issue #1568, opened 2019-07-20
v4: plugin feature request: minify template literals


vegarringdal:
Started using lit-html lately.
Would be great if fusebox could minify template literals

Found this, maybe it can help you make a plugin after v4 is released
https://github.com/asyncLiz/minify-html-literals#readme

kyr0:
You can solve this using `customTransformers` now (recently added :) Just scan the AST for template literals and return the minified text. The modified AST will be given to the typescript compiler. You can also write a plugin that registers this transformer in the context config, so when your plugin is used, the customTransformer is used.

vegarringdal:
@kyr0 
Nice, will try and look at this in the weekend.
Might ask you for help, since I have really no clue about this :-)

vegarringdal:
> You can also write a plugin that registers this transformer in the context config, so when your plugin is used, the customTransformer is used.

@kyr0 
Wont anything I add be destroyed by the plugin,or where you thinking about something else ?

https://github.com/fuse-box/fuse-box/blob/master/src/plugins/core/plugin_customtransform.ts
Maybe this should check if there is allready added customTransformers ?

vegarringdal:
fixed

PR #1623, opened 2019-09-07
feat: pluginMinifyHtmlLiterals #1568

vegarringdal:
@nchanged @kyr0 

Have tried implement #1568.
Hopefully I havent messed up to much :joy:

### DIFF
--- a/src/plugins/core/plugin_minify_html_literals.ts
+++ b/src/plugins/core/plugin_minify_html_literals.ts
@@ -1,12 +1,31 @@
-import { minifyHTMLLiterals } from 'minify-html-literals';
 import { Context } from '../../core/Context';
 
-export function pluginMinifyHtmlLiterals(...options: any) {
+/**
+ * See https://github.com/asyncLiz/minify-html-literals after source input
+ */
+interface IPluginMinifyHtmlLiterals {}
+
+/**
+ * Simple plugin to use npm module minify-html-literals
+ * @param options
+ */
+export function pluginMinifyHtmlLiterals(...options: IPluginMinifyHtmlLiterals[]) {
   const useroptions = options;
+
   return (ctx: Context) => {
-    ctx.ict.on('assemble_fast_analysis', props => {
+    ctx.ict.on('bundle_resolve_module', props => {
       const { module } = props;
-      if (module.props && (module.props.extension === '.ts' || module.props.extension === '.js')) {
+      if (module.isExecutable()) {
+        // verify if module is installed
+        if (!ctx.isInstalled('minify-html-literals')) {
+          ctx.fatal(`Fatal error when capturing ${module.props.absPath}`, [
+            'Module "minify-html-literals" is required, Please install it using the following command',
+            'npm install minify-html-literals --save-dev',
+          ]);
+          return;
+        }
+
+        // check if we have any html`` or css`` template tags
         if (module.contents.indexOf('html`') !== -1 || module.contents.indexOf('css`') !== -1) {
           ctx.log.info('pluginMinifyHtmlLiterals', 'minifying $file', {
             file: module.props.absPath,
@@ -15,6 +34,7 @@ export function pluginMinifyHtmlLiterals(...options: any) {
           // https://github.com/asyncLiz/minify-html-literals
           let newcontent;
           try {
+            const minifyHTMLLiterals = require('minify-html-literals');
             newcontent = minifyHTMLLiterals(module.contents, ...useroptions);
           } catch {}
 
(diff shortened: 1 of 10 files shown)


src/plugins/core/plugin_minify_html_literals.ts
```diff
@@ -0,0 +1,33 @@
+import { minifyHTMLLiterals } from 'minify-html-literals';
+import { Context } from '../../core/Context';
+
+export function pluginMinifyHtmlLiterals(...options: any) {
+  const useroptions = options;
+  return (ctx: Context) => {
+    ctx.ict.on('assemble_fast_analysis', props => {
+      const { module } = props;
+      if (module.props && (module.props.extension === '.ts' || module.props.extension === '.js')) {
```
nchanged: `module.isExecutable()` ;-)

src/plugins/core/plugin_minify_html_literals.ts
```diff
@@ -0,0 +1,33 @@
...
+        if (module.contents.indexOf('html`') !== -1 || module.contents.indexOf('css`') !== -1) {
```
nchanged: we can remove that probably

src/plugins/core/plugin_minify_html_literals.ts
```diff
@@ -0,0 +1,33 @@
+import { minifyHTMLLiterals } from 'minify-html-literals';
+import { Context } from '../../core/Context';
+
+export function pluginMinifyHtmlLiterals(...options: any) {
```
nchanged: can you add an interface IPluginMinifyHtmlLiterals to follow the conventions?

src/plugins/core/plugin_minify_html_literals.ts
```diff
@@ -0,0 +1,33 @@
+import { minifyHTMLLiterals } from 'minify-html-literals';
```
nchanged: don't import it here ;_)
do it like this: 
https://github.com/fuse-box/fuse-box/blob/master/src/plugins/core/plugin_less.ts#L11-L17

cuz that package is in dev dependencies it will simply fail

src/plugins/core/plugin_minify_html_literals.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Context } from '../../core/Context';
+
+export function pluginMinifyHtmlLiterals(...options: any) {
+  const useroptions = options;
+  return (ctx: Context) => {
+    ctx.ict.on('assemble_fast_analysis', props => {
```
nchanged: I also think it should be at `bundle_resolve_module` level, `assemble_fast_analysis` is a bit too early

src/plugins/core/plugin_minify_html_literals.ts
vegarringdal: Very nice, will update

src/plugins/core/plugin_minify_html_literals.ts
vegarringdal: I figured it was best to check first. If there is not template tags with those names we dont need to run the plugin, indexof will be faster then using the plugin. 
Want to keep fusebox fast :-)

src/plugins/core/plugin_minify_html_literals.ts
vegarringdal: ok.
just do `require('minify-html-literals')` before I use it then ?

src/plugins/core/plugin_minify_html_literals.ts
vegarringdal: ok, I really dont know what comes first and last atm.
So will just update like you say :-)

src/plugins/core/plugin_minify_html_literals.ts
vegarringdal: yes

vegarringdal:
Implemented the review comments.
keept the check if html/css template strings only. will be faster I think.

image

added quickfix for thsi for now with any.
`cssBundle.override((response as any).styles, (response as any).sourceMap.toString());`

vegarringdal:
@nchanged added what we talked about. have a last look :-)

src/plugins/core/plugin_minify_html_literals.ts
```diff
@@ -0,0 +1,53 @@
...
+          ctx.log.info('pluginMinifyHtmlLiterals', 'minifying $file', {
+            file: module.props.absPath,
+          });
+
+          // https://github.com/asyncLiz/minify-html-literals
+          let newcontent;
+          try {
+            const minifyHTMLLiterals = require('minify-html-literals');
+            newcontent = minifyHTMLLiterals(module.contents, ...useroptions);
+          } catch {}
```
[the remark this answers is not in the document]
vegarringdal: Guessing you saw I did a error if it did not generate code since its merged :-)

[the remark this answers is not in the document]
nchanged: yeah but still it's a bit weird tho ;-)

PR #1624, opened 2019-09-07
fix: give ts:main precedence over module in package.json

vegarringdal:
Tried to fix the order.
ts_main should come before module/main but after local:main (alias/paths directories)


vegarringdal:
Yes, I will add some tests.

vegarringdal:
Tests added

Issue #1618, opened 2019-09-06
Generate index.html if file not found when using webIndex


nchanged:
Generate index.html on the specific path (`src/index.html`) if file not found when using webIndex with default contents (which are available in the repository)

@kyr0  since you been tinkering with webIndex anyway, do you think you can take that one in to?

kyr0:
Read it too late, but I'll do this in another PR then, it's simple.

kyr0:
@nchanged Close? :)

PR #1626, opened 2019-09-08
feat(WebIndexDefaultTpl): Default WebIndex template allows for zero-config setup

kyr0:
Fixes issue #1618 -- but I took a bit different approach, open for discussion:

1) I didn't generate the index.html because we'd change the developers code by doing this. Thats always a dangerous thing to do for a build system, especially when we do that automatically -- instead I use a default template on the fly and warn that there is no index template file. So if the developer wants, the file can be created by him but if he doesn't, the project builds anyway. Perfect for zero-config setups and super simple setups.

2) I changed the default template to add some useful headers for modern devices like viewport and edge support, UTF-8 charset etc.

### DIFF
--- a/src/web-index/webIndex.ts
+++ b/src/web-index/webIndex.ts
@@ -4,6 +4,7 @@ import { Context } from '../core/Context';
 import { env } from '../env';
 import { ensureAbsolutePath, fileExists, readFile, joinFuseBoxPath } from '../utils/utils';
 import { htmlStrings } from './htmlStrings';
+import { FuseBoxLogAdapter } from '../fuse-log/FuseBoxLogAdapter';
 
 export interface IWebIndexConfig {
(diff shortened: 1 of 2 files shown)


src/web-index/webIndex.ts
```diff
@@ -26,16 +44,14 @@ export function replaceWebIndexStrings(str: string, keys: { [key: string]: any }
   });
 }
 
-export function getEssentialWebIndexParams(config: IWebIndexConfig | boolean) {
+export function getEssentialWebIndexParams(config: IWebIndexConfig | boolean, log: FuseBoxLogAdapter) {
   let templatePath = join(env.FUSE_MODULES, 'web-index-default-template/template.html');
```
nchanged: should we use the contents of this default template instead of the constant you defined earlier?

Issue #1582, opened 2019-08-17
V4- sparky src/dist - bug ?


vegarringdal:
If I have a sparky task script file inside folders it does not work as expected.
Best behavior would be if it used scripts location as basepath for src and dest.

Ready to test version of script is here
https://github.com/vegarringdal/bug-4
scripts
* `npm run copy-bug1`
* `npm run copy-bug2`
* `npm run copy-hack`

```js
const PATH = process.argv[3];
if (PATH) {
  process.chdir(__dirname);
}

const { sparky } = require("fuse-box");
class Context {}
const { task, src } = sparky(Context);

task("copy-bug1", async context => {
  // expected behavior would be that this uses script location "_dirname" as basepath on both
  // with this one nothing happend because it does not find any files
  await src(`./src/**/*.*`)
    .dest(`./dist/ts`, `src`)
    .exec();
});

task("copy-bug2", async context => {
    // this will find files
    // but it will break and say it cant find file under :
    // .... package/cool/package/cool/src/index.ts
    // looks like it adds _dirname to the src path
    await src(`./package/cool/src/**/*.*`)
      .dest(`${PATH}/dist/ts`, `src`)
      .exec();
  });

task("copy-hack", async context => {
  await src(`./src/**/*.*`)
    // supply it with path './package/cool'
    .dest(`${PATH}/dist/ts`, `src`)
    .exec();
});


```

vegarringdal:
closing

nchanged:
@vegarringdal sorry did you get it resolved? ;-)

vegarringdal:
I though I just messed up so removed it.
But now when I test more it looks like something goes very wrong when my file with the task in not on root where npm gets run from

vegarringdal:
or Im just very tired messing up

vegarringdal:
@nchanged 
Is this understandable ?

```js

// script gets started from "npm start"
// this runs : node ./package/cool/tasks copy-now


const { sparky } = require("fuse-box");
class Context {}
const { task, src } = sparky(Context);

// unless I have this it fails
process.chdir(__dirname);


task("copy-now", async context => {
  await src(`./src/**/*.*`)
    .dest('./package/cool/dist/ts', `/src/`) 
    // the dest() is not behaving like I expected, its like its not following process.cwd() or __dirname
    // but unless I set process.cwd() to __dirname it wont copy, but I still haveto use npm base ?
    .exec();
});

```

vegarringdal:
[...]

PR #1628, opened 2019-09-09
fix: sparky path executed in sub folders #1582

vegarringdal:
@nchanged 

Think Ive found the root of my issue.
Need to test more and find a good way to add test to this. I only ran the sparky tests. 

Problem is when its started from package root with something like npm or directly `node ./projects/sample1/fuse.ts`

If I tried to run this nothing happend, because it didnt find it. 
This is why I added option cwd to GLOB.
Changed the ROOT app since files didnt end up relative to script using sparky.

```
// failing, but dont know why
task("copy-src", async context => {
  await src(`./src/**/*.*`)
    .dest('./dist3/ts', `/src/`) 
    .exec();
});
```

If Ive broken something I dont know atm.
But like I started, we need to add more tests before we merge.
Maybe CI finds something when I create the PR too 😂 
Its late, I need sleep now.

### DIFF
--- a/src/sparky/sparky_chain.ts
+++ b/src/sparky/sparky_chain.ts
@@ -37,7 +37,7 @@ export function sparkyChain(log: FuseBoxLogAdapter): ISparkyChain {
     }
 
     if (newLocation && newLocationBase) {
-      const root = ensureAbsolutePath(newLocation, env.APP_ROOT);
+      const root = ensureAbsolutePath(newLocation, env.SCRIPT_PATH);
       for (const i in latest) {
         const file = latest[i];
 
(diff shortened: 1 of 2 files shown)


src/sparky/sparky_chain.ts
```diff
@@ -37,7 +37,7 @@ export function sparkyChain(log: FuseBoxLogAdapter): ISparkyChain {
     }
 
     if (newLocation && newLocationBase) {
-      const root = ensureAbsolutePath(newLocation, env.APP_ROOT);
+      const root = ensureAbsolutePath(newLocation, env.SCRIPT_PATH);
```
nchanged: Yeah I think that's what It should be.. I never had problems cuz fuse.ts is usually at the root ;-)

PR #1629, opened 2019-09-10
fix: sparky paths on windows

vegarringdal:
More fixes.
Ive never tried to build on window, so this error have always been here.

**Update: removed part with test failing on webindex etc, Ill do that in own PR.**


vegarringdal:
Had a new look

Looks like this function does not play very nice to windows paths
```ts
export function path2RegexPattern(path: string) {
  path = path
    .split('.')
    .join('\\.')
    .split('/')
    .join('\\/');
  return new RegExp(path);
}
```

vegarringdal:
@nchanged 

I also added the fix to tests here, since its also related to window only.
Important you have a look at this part, regex isnt something Im good at.
The split/join part works very badly when it also needed to do handle windows paths _(I could not wrap my brain around it, just broke something else when I fix one thing)_
So reused some other function that pretty much did the same just better.

src/plugins/pluginUtils.ts
```diff
@@ -22,16 +22,20 @@ export function simplifiedRegExp(input: undefined | string | RegExp): RegExp {
...
     return;
   }
+
   if (typeof input === 'string') {
     let r = '';
     for (let i = 0; i < input.length; i++) {
       switch (input[i]) {
         case '.':
-          r += '\\.';
+          r += '(\\.|.)';
```
nchanged: that doesn't make sense ;-) We need to escape all the periods, otherwise it will become a part of Regex

vegarringdal: fixed, we also need to test sparky better.
maybe manually spawing a child process and make it run a task ?

src/plugins/pluginUtils.ts
```diff
@@ -32,6 +33,9 @@ export function simplifiedRegExp(input: undefined | string | RegExp): RegExp {
         case '/':
           r += '(\\/|\\\\)';
           break;
+        case '\\':
```
nchanged: so that's when a absolute path on windows is passed right?

src/utils/utils.ts
```diff
@@ -165,12 +166,7 @@ export function findReplace(str: string, re: RegExp, fn: (args) => string) {
   });
 }
 export function path2RegexPattern(path: string) {
```
nchanged: can we just get rid of that function and use simplifiedRegex?

src/plugins/pluginUtils.ts
nchanged: do you think you can give a test? that's quite an important part of the process

vegarringdal: > so that's when a absolute path on windows is passed right?

All paths have `\\` while linux/max uses `/`.
Here is a typical path from my vmware, the `\` will ofc be escaped.. so really `\\`
`C:\Users\Developer\Desktop\New folder\plugins\core`
This is why I removed the split/join part and just used this.


> do you think you can give a test? that's quite an important part of the process

I will try and give this and sparky some tests.
Limted time this week and not very good at tests, so its just going to a little while :-)
But will try and improve 👍 
This is under plugin utils, but all path matching prob should be move to 1 place. so no-mather what you do we all use same tested functions.
Hopyfully Im the only one using npm scripts in subfolders and windows 😂

src/utils/utils.ts
vegarringdal: Ill have a look

Do we want to change back to using split/join likei was before too?
Might work if I give it another try, dunno if it willbe faster, just need to do the windows paths before anything else

vegarringdal: I currently replaced all `simplifiedRegex` to `path2RegexPattern`
`simplifiedRegex` does tell me anythin, while `path2RegexPattern` is more spesific in what it does
`path2RegexPattern` is only located under utils now.

vegarringdal:
@nchanged 

Summary:
* `path2RegexPattern` replaced simplified regex all places I found it, this name said what it actually do.
* Fix a bug in path2RegexPattern, need to use same pattern as unix. so when testing path it does not matter if they have used unix/window type.
* I added some tests for sparky and path2RegexPattern. Not the best test, but show sub folders work.

See test still work.
Very tried atm, so nice if you give it a new review/check.

src/plugins/pluginUtils.ts
```diff
@@ -14,32 +14,6 @@ export function testPath(input: string, target: string | RegExp) {
 
 export function parsePluginOptions<T>(a, b?, defaultValue?): [T, RegExp] {
   const opts = b ? b : typeof a === 'object' ? a : defaultValue;
-  const rex = b || typeof a === 'string' || isRegExp(a) ? simplifiedRegExp(a) : undefined;
+  const rex = b || typeof a === 'string' || isRegExp(a) ? path2RegexPattern(a) : undefined;
   return [opts, rex];
 }
-
-export function simplifiedRegExp(input: undefined | string | RegExp): RegExp {
```
nchanged: Why did u remove this function ?))

Issue #1641, opened 2019-09-21
Code splitting cannot handle paths larger than 2 deep


patrickmichalina:
It outputs the bundles, but truncates the filename to something like `026a6b27f-.js`

Example input `  import('./site/about/about.module').then(m => m.AboutModule) `

patrickmichalina:
Fixed in https://github.com/fuse-box/fuse-box/pull/1642

PR #1642, opened 2019-09-21
fix: beautifyBundleName handles arbitrary length

patrickmichalina:
### Improved configuration for code splitting

- Allow for removal of file hash from path
- Allow for configuring max filename output length
- Improved logic for pathname creation (see: https://github.com/fuse-box/fuse-box/issues/1641)

```js
fusebox({
  codeSplitting: {
    useHash: true | false // default: true,
    maxPathLength: 200 // default: 20
 }
})

```


package.json
```diff
@@ -141,6 +141,7 @@
     "format:fix": "prettier --write \"src/**/*.{js,json,md,ts,yml}\"",
     "document": "ts-node -T fuse document",
     "test": "jest --verbose",
+    "test.watch": "jest --watch --verbose",
```
patrickmichalina: Easier to run tests when developing :)

src/production/transformation/__tests__/__snapshots__/condition.unwrapper.test.ts.shapshot.json
```diff
@@ -45,5 +45,44 @@
   "41d36129": "alert('no')",
   "573e3511": "alert('no')",
   "07044b600": "console.log(\"yes\")",
-  "f91ebe7": "function test(){\n          return;\n      }\n      "
+  "f91ebe7": "function test(){\n          return;\n      }\n      ",
```
nchanged: what's this? ;-) why has it been updated?

src/utils/__tests__/utils.test.ts
```diff
@@ -19,12 +19,30 @@ describe('utils', () => {
...
     });
     it('should beautifyBundleName 2', () => {
       const name = beautifyBundleName('oi.js');
       expect(name).toEqual('oi');
     });
+    it('should beautifyBundleName 3', () => {
+      const name = beautifyBundleName('./not-found-page/not-found-page/not-found-page.module');
+      expect(name).toEqual('not-found-page');
+    });
+    it('should beautifyBundleName 4', () => {
```
nchanged: I think it should have a limit, it can't be longer than... 20 symbols?

patrickmichalina: These files are Jest snapshot files. Automatically occurred when I added tests....

patrickmichalina: Why is there a limit at all? I kept the default behavior but fixed the case that reduce the string to empty!

patrickmichalina: Sorta, the change is subtle. It exceeds 20 characters, then it will use the last value in the path split. The last value in the array (even if that exceeds the limit)

src/production/transformation/__tests__/__snapshots__/condition.unwrapper.test.ts.shapshot.json
patrickmichalina: These files are Jest snapshot files. Automatically occurred when I added tests....

patrickmichalina: We can turn this off in Jest config. Are you familiar with Jest Snapshots feature? https://jestjs.io/docs/en/snapshot-testing

More used for UI code.

src/production/stages/codeSplittingStage.ts
```diff
@@ -45,9 +45,12 @@ export function codeSplittingStage(props: IProductionFlow) {
     processEntry(props, link);
 
     if (root.dynamicDependencies.length > 0) {
+      const useHash = props.ctx.config.codeSplitting.useHash
```
nchanged: could you make it `true` if it's undefined?

patrickmichalina: done

PR #1643, opened 2019-09-21
fix(code-splitting): pass correct config value + other changes

patrickmichalina:
### Improved output

Screen Shot 2019-09-21 at 2 01 28 PM

Notice `"ngc-browser"`, that is the entry to the project files being compiled. I do think it would be nice to be able to remove "ngc-browser" (or whatever in a given project) by exposing an option in config, or if there is a way to automatically detect this in the fusebox context I am not aware of.

### DIFF
--- a/src/config/PrivateConfig.ts
+++ b/src/config/PrivateConfig.ts
@@ -118,8 +118,8 @@ export class PrivateConfig {
     }
     // define default settings for code splitting
     this.codeSplitting = props.codeSplitting || {};
-    this.codeSplitting.useHash = typeof this.codeSplitting.useHash === undefined ? true : this.codeSplitting.useHash
-    this.codeSplitting.maxPathLength = this.codeSplitting.maxPathLength || 20
+    this.codeSplitting.useHash = typeof this.codeSplitting.useHash === undefined ? true : this.codeSplitting.useHash;
+    this.codeSplitting.maxPathLength = typeof this.codeSplitting.maxPathLength === 'number' ? this.codeSplitting.maxPathLength : 20;
 
     if (!this.codeSplitting.scriptRoot) {
       if (this.isServer()) {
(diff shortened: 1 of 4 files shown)


src/config/PrivateConfig.ts
```diff
@@ -118,8 +118,8 @@ export class PrivateConfig {
     }
     // define default settings for code splitting
     this.codeSplitting = props.codeSplitting || {};
-    this.codeSplitting.useHash = typeof this.codeSplitting.useHash === undefined ? true : this.codeSplitting.useHash
-    this.codeSplitting.maxPathLength = this.codeSplitting.maxPathLength || 20
+    this.codeSplitting.useHash = typeof this.codeSplitting.useHash === undefined ? true : this.codeSplitting.useHash;
+    this.codeSplitting.maxPathLength = typeof this.codeSplitting.maxPathLength === 'number' ? this.codeSplitting.maxPathLength : 20;
```
patrickmichalina: Allows `0` values through.

src/production/stages/codeSplittingStage.ts
```diff
@@ -50,7 +50,7 @@ export function codeSplittingStage(props: IProductionFlow) {
 
       const bundle = createBundle({
         ctx: props.ctx,
-        name: `${useHash ? fastHash(root.module.props.absPath) + '-' : ''}${beautifyBundleName(root.module.props.absPath, maxlen)}`,
+        name: `${useHash ? fastHash(root.module.props.absPath) + '-' : ''}${beautifyBundleName(root.module.props.fuseBoxPath, maxlen)}`,
```
patrickmichalina: `fuseBoxPath` is a nicer reference that doesn't include the larger absolute path. This gives us "hash like" structure to reduce filename collision.

nchanged: I think you should include pkg.getPublicName which might Include version. Or at least check if there are external deps (basically you are able to lazy load a library - that's my concern)

patrickmichalina: absPath does this but fuseBoxPath doesn’t?

Are you saying for the case 


Import(‘https://lodash.com”)

?

patrickmichalina: Do you have a test-case example?

PR #1650, opened 2019-09-27
Fix web index

kyr0:
- publicPath handles "" and "./" correctly
- webIndex plugins returns of bundles, cssTags, scriptTags is respected and actually used

Bildschirmfoto 2019-09-28 um 01 26 38
Bildschirmfoto 2019-09-28 um 01 26 28

### DIFF
--- a/src/config/PrivateConfig.ts
+++ b/src/config/PrivateConfig.ts
@@ -119,20 +119,20 @@ export class PrivateConfig {
     // define default settings for code splitting
     this.codeSplitting = props.codeSplitting || {};
     this.codeSplitting.useHash = typeof this.codeSplitting.useHash === undefined ? true : this.codeSplitting.useHash;
-    this.codeSplitting.maxPathLength = typeof this.codeSplitting.maxPathLength === 'number' ? this.codeSplitting.maxPathLength : 20;
+    this.codeSplitting.maxPathLength =
+      typeof this.codeSplitting.maxPathLength === 'number' ? this.codeSplitting.maxPathLength : 20;
 
     if (!this.codeSplitting.scriptRoot) {
       if (this.isServer()) {
         this.codeSplitting.scriptRoot = './';
       } else {
-        if (this.webIndex && this.webIndex.publicPath) {
+        if (this.webIndex && typeof this.webIndex.publicPath != 'undefined') {
           this.codeSplitting.scriptRoot = this.webIndex.publicPath;
         } else {
           this.codeSplitting.scriptRoot = '/';
         }
       }
     }
-
 
     this.watch = {
       enabled: !env.isTest,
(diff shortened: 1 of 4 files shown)


src/utils/utils.ts
```diff
@@ -244,9 +242,13 @@ export function ensureFuseBoxPath(input: string) {
 
 export function joinFuseBoxPath(...any): string {
   let includesProtocol = any[0].includes('://');
+  let isRelative = any[0].includes('./');
   let joinedPath = !includesProtocol
     ? path.join(...any)
     : any[0].replace(/([^/])$/, '$1/') + path.join(...any.slice(1));
+  if (isRelative) {
+    joinedPath = './' + joinedPath;
```
nchanged: Oh oh... You can't do that. That will break a lot of things, starting from the client side
`__file("hello/foo.bar")` will become `__file("./hello/foo.bar")` and everything will get broken

kyr0: So, what's your idea on solving this? :) we could fix this only for the use of webIndex etc. on the return side of the call

src/config/PrivateConfig.ts
```diff
@@ -119,21 +119,21 @@ export class PrivateConfig {
...
-    this.codeSplitting.maxPathLength = typeof this.codeSplitting.maxPathLength === 'number' ? this.codeSplitting.maxPathLength : 20;
+    this.codeSplitting.maxPathLength =
+      typeof this.codeSplitting.maxPathLength === 'number' ? this.codeSplitting.maxPathLength : 20;
 
     if (!this.codeSplitting.scriptRoot) {
       if (this.isServer()) {
         this.codeSplitting.scriptRoot = './';
       } else {
-        if (this.webIndex && this.webIndex.publicPath) {
+        if (this.webIndex && typeof this.webIndex.publicPath != 'undefined') {
```
patrickmichalina:
```suggestion
        if (this.webIndex && typeof this.webIndex.publicPath === 'string') {
```

src/config/PrivateConfig.ts
```diff
@@ -204,10 +204,10 @@ export class PrivateConfig {
       return userPublicPath;
     }
     let publicPath = '/';
-    if (this.webIndex && this.webIndex.publicPath) {
+    if (this.webIndex && typeof this.webIndex.publicPath != 'undefined') {
```
patrickmichalina:
```suggestion
    if (this.webIndex && typeof this.webIndex.publicPath === 'string') {
```

src/config/PrivateConfig.ts
```diff
@@ -204,10 +204,10 @@ export class PrivateConfig {
...
       publicPath = this.webIndex.publicPath;
     }
-    if (this.resources && this.resources.resourcePublicRoot) {
+    if (this.resources && typeof this.resources.resourcePublicRoot != 'undefined') {
```
patrickmichalina:
```suggestion
    if (this.resources && typeof this.resources.resourcePublicRoot === 'string') {
```

src/config/PrivateConfig.ts
nchanged: Agreed )

PR #1655, opened 2019-10-04
Add options passed to less compiler.

ldom66:
I would like to pass options to the less compiler such as { modifyVars: { 'accent-color': 'blue' } }.

I propose to add an options object to stylesheet.less that would be passed to less.render().

### DIFF
--- a/src/config/ILessProps.ts
+++ b/src/config/ILessProps.ts
@@ -1,3 +1,4 @@
 export interface ILessProps {
   plugins: Array<any>;
+  options: any;
 }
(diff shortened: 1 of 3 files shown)


src/config/ILessProps.ts
```diff
@@ -1,3 +1,4 @@
 export interface ILessProps {
   plugins: Array<any>;
+  options: any;
```
nchanged: Can you make typings for this?

nchanged:
@ldom66  looks good!! Can I ask you to update the documentation too?

src/config/ILessProps.ts
ldom66: We could use Less typings https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/less/index.d.ts
But how would I import typings for another project?

PR #1691, opened 2019-11-26
Add default matcher to pluginLess

Marcisbee:
This PR fixes this issue: https://github.com/fuse-box/angular2-example/pull/36

Problem was that LESS plugin failed if no matcher was defined (like in docs), so I guess we would create a default matcher like in SASS plugin for this to work:
```ts
pluginLess({ asText: true, useDefault: false })
```

Altho the workaround is to use matcher
```ts
pluginLess('*.less', { asText: true, useDefault: false })
```

I think this should be fixed as it was possible to define plugin without a matcher and then it's just a potato that does nothing.


src/frameworks/angular/__tests__/__snapshots__/angularURLReplacer.test.ts.shapshot.json
```diff
@@ -1,4 +1,6 @@
 {
   "0deb0dd1": "\n@Component({\n  selector: \"app-root\",\n  template : require(\"./hello.html\"),\n  styles : [require(\"./foo.css\"),require(\"bar.scss\")],\n})\n",
-  "0deb0dd0": "\n@Component({\n  selector: \"app-root\",\n  template : require(\"./hello.html\"),\n  styles : [require(\"bar.scss\")],\n})\n"
+  "0deb0dd0": "\n@Component({\n  selector: \"app-root\",\n  template : require(\"./hello.html\"),\n  styles : [require(\"bar.scss\")],\n})\n",
+  "04900d8c": "\n@Component({\n  selector: \"app-root\",\n  template : require(\"./hello.html\"),\n  styles : [require(\"./foo.css\"),require(\"bar.scss\")],\n})\n",
+  "04900d8b": "\n@Component({\n  selector: \"app-root\",\n  template : require(\"./hello.html\"),\n  styles : [require(\"bar.scss\")],\n})\n"
```
nchanged: Could you discard those changes? It looks like the wrong hash is generated

Marcisbee: Sure

nchanged:
👍 thanks a million!

PR #1695, opened 2019-12-07
More doc updates

SephReed:
Added a full-config.md page.  Could really use some help listing all the default values for items that are not undefined or false by default.


SephReed:
This concludes my full parsing of the entirety of the docs.  Added a few new pages, mostly just styled/restructured things (bullet points, numbered items, page breaks).  Changed present tense to future tense.

Issue #1690, opened 2019-11-21
"Smarter" suggestions when plugin dependency is missing


paxperscientiam:
This is an easy one for anyone else who wants to pick it up, but I'm happy to also.

Looking at the sass plugin, i saw this error message that is given when node-sass is not installed in the project:
```
'Module "sass" is required, Please install it using the following command',
          'npm install node-sass --save-dev',
```
Would be neat if, as part of environment detection, the likely package manager could be identitfied.

presence of .yarnrc => yarn
presence of pnpm-lock.yaml => pnpm
presence of package-lock.json => npm
We could then set some variable PKG_MANAGER accordingly.

So. `'npm install node-sass --save-dev'` becomes ``${PKG_MANAGER} install node-sass --save-dev``

nchanged:
@paxperscientiam sounds like a great idea, however not the biggest priority ATM. Do you want to take a look and make a PR?

paxperscientiam:
@nchanged yes!

charsleysa:
Closed due to inactivity.

PR #1701, opened 2019-12-11
feat: Check the package manager

paxperscientiam:
So, this PR begins to address #1690. Checks if user is using yarn, then pnpm, and then falls to npm. What do you think about this heuristic -- too simple?

I noticed that a Sparky task uses npm. Would you want me to address that in this PR or another one?


SephReed:
I don't know yarn (or know if you know it).  Here's a quick link I found on syntax differences between it and npm:

https://stackoverflow.com/a/40137938/4808079

paxperscientiam:
@SephReed thanks dude. I've used yarn lots, but was a bit hasty. I'll fix

src/env.ts
```diff
@@ -19,6 +20,24 @@ export const env = {
...
+  }
+
+  if (fs.existsSync(path.join(FUSE_ROOT, './.yarnrc'))
+      || fs.existsSync(path.join(FUSE_ROOT, './yarn.lock'))) {
+    return new Data('yarn', 'yarn add', 'yarn add --dev');
+  } else if (fs.existsSync(path.join(FUSE_ROOT, './pnpm-lock.yaml'))) {
+    return new Data('pnpm', 'pnpm add', 'pnpm add --save-dev');
+  } else {
+    // package-lock.json
+    return new Data('npm', 'npm install', 'npm install --save-dev');;
```
nchanged: can't you use just return `{name : "npm", ...}`
Overall it looks good, but I don't like this class here ;-)

paxperscientiam: fixed

src/env.ts
```diff
@@ -21,20 +21,16 @@ export const env = {
 };
 
 export function getPackageManagerData() {
-  function Data(name: string, installCmd: string, installDevCmd: string) {
-    this.name = name;
-    this.installCmd = installCmd;
-    this.installDevCmd = installDevCmd;
-  }
-
+  console.log("SHIT");
```
SephReed: Heh.  Maybe take it out though.

paxperscientiam:
@nchanged Not really sure what to do with conflicting package-lock.json and package.json

src/env.ts
paxperscientiam: done lol

nchanged:
@paxperscientiam could you re-open the PR againts `grand-refactor` branch?

PR #1702, opened 2019-12-11
Feat -  Find up and monorepo fixes

SephReed:
Monorepo setup was unable to find my package.json because it was not at the entry point of the project src.  Added a new util which searches upwards for files.

### DIFF
--- a/src/resolver/fileLookup.ts
+++ b/src/resolver/fileLookup.ts
@@ -2,9 +2,7 @@ import * as path from 'path';
 import * as fs from 'fs';
 import { getFolderEntryPointFromPackageJSON } from './shared';
 import { TypescriptConfig } from '../interfaces/TypescriptInterfaces';
-import { initTypescriptConfig } from '../tsconfig/configParser';
 import { fileExists } from '../utils/utils';
-import { findUp } from '../utils/findUp';
 
 export interface ILookupProps {
   typescriptFirst?: boolean;
@@ -42,24 +40,6 @@ const TS_EXTENSIONS = ['.ts', '.tsx'];
 const TS_EXTENSIONS_FIRST = [...TS_EXTENSIONS, ...JS_EXTENSIONS];
 const JS_EXTENSIONS_FIRST = [...JS_EXTENSIONS, ...TS_EXTENSIONS];
 
-function tryIndexes(target: string, indexes: Array<string>) {
-  for (const i in indexes) {
-    const indexFile = indexes[i];
-    const resolved = path.join(target, indexFile);
-    if (fileExists(resolved)) {
-      return resolved;
-    }
-  }
-}
-
-function tryExtensions(target: string, extensions: Array<string>) {
-  for (const i in extensions) {
-    const resolved = `${target}${extensions[i]}`;
-    if (fileExists(resolved)) {
-      return resolved;
-    }
-  }
-}
 
 export function fileLookup(props: ILookupProps): ILookupResult {
   if (!props.fileDir && !props.filePath) {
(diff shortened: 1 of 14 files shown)


src/resolver/fileLookup.ts
```diff
@@ -106,24 +107,32 @@ export function fileLookup(props: ILookupProps): ILookupResult {
       let monorepoModulesPaths;
       let tsConfigAtPath: TsConfigAtPath;
 
+
       // only in case of a directory
-      const packageJSONPath = path.join(resolved, 'package.json');
+      const projectRoot = path.join(resolved, "../../../"); // TODO: get the actual projectRoot
```
SephReed: Didn't know a good way to get fusebox's homeDir

src/resolver/fileLookup.ts
```diff
@@ -106,24 +107,32 @@ export function fileLookup(props: ILookupProps): ILookupResult {
...
 
         if (useLocalMain && packageJSON['local:main']) {
-          const _monoModules = path.resolve(resolved, 'node_modules');
+          const _monoModules = path.resolve(path.dirname(packageJSONPath), 'node_modules');
           if (fileExists(_monoModules)) {
             monorepoModulesPaths = _monoModules;
           }
 
-          const _tsConfig = path.resolve(resolved, 'tsconfig.json');
+          const _tsConfig = findUp(resolved, 'tsconfig.json', {
```
SephReed: Can a subproject use the monorepos tsconfig?  Should it be `inclusive: false`

src/resolver/fileLookup.ts
```diff
@@ -106,24 +107,32 @@ export function fileLookup(props: ILookupProps): ILookupResult {
...
+            boundary: projectRoot,
+            inclusive: true,
+          });
```
SephReed: Can a subproject use the monorepos tsconfig?  Should it be `inclusive: false`

src/resolver/fileLookup.ts
```diff
@@ -100,43 +102,56 @@ export function fileLookup(props: ILookupProps): ILookupResult {
...
+          boundary: projectRoot,
+          inclusive: false,
+        });
+        const packageRoot = path.dirname(packageJSONPath)
+        if (fileExists(packageJSONPath)) {
+          const useLocalMain = !/node_modules/.test(packageJSONPath);
+          const packageJSON = require(packageJSONPath);
+          const entry = getFolderEntryPointFromPackageJSON({ json: packageJSON, useLocalField: useLocalMain });
+
+          if (packageJSON['main:local']) {
```
nchanged: Do we need this check? xD ;-))

src/resolver/nodeModuleLookup.ts
```diff
@@ -92,7 +109,24 @@ export interface INodeModuleLookup {
 }
 
 export function nodeModuleLookup(props: IResolverProps, parsed: IModuleParsed): INodeModuleLookup {
-  const folder = findTargetFolder(props, parsed);
+  let folder = findTargetFolder(props, parsed);
+  let packageJSONFile: string;
+  let useLocalField = false;
+  if (!folder) {
+    folder = findTargetFolderInTypescriptPaths(props);
```
nchanged: I think this isn't gonna work, it will never hit it. Since "local:main" is used against local module and never against the real ones. Since monorepo structure ends up being like a normal projects with replaced paths that look like packages but reference local files instead. We're adding an additional overhead here.

src/resolver/fileLookup.ts
```diff
@@ -100,43 +102,56 @@ export function fileLookup(props: ILookupProps): ILookupResult {
...
-          if (fileExists(_monoModules)) {
-            monorepoModulesPaths = _monoModules;
+    isDirectory = stat.isDirectory();
+
+    if (isDirectory) {
+      if (props.inBaseUrl !== true) {
+        let monorepoModulesPaths;
+        let tsConfigAtPath: TsConfigAtPath;
+        // only in case of a directory
+        const projectRoot = path.join(resolved, "../../../"); // TODO: get the actual projectRoot
```
nchanged: This doesn't look right ;-)

src/resolver/fileLookup.ts
SephReed: Silly as it is, I ended up banging my head for a while before I realized I'd inverted the order of these keywords.

I'd rather find a way to check if something really looks like it should be a "sibling module" and give a small notice if there is no "local:main".

src/resolver/nodeModuleLookup.ts
SephReed: I wrote this as I ran into errors.  It did end up fixing the issue, but now that I understand the repo more, it's out of place.  I think "sibling modules" should probably be found after attempting pathLookup and before moduleLookup.

Or perhaps they should be part of pathLookup?

src/resolver/fileLookup.ts
SephReed: Yeah.  I'm beginning to  think figuring out whether or not something is a module shouldn't be done inside fileLookup.  Maybe paths lookup would be better...  Not sure.  But all the other lookups eventually reference fileLookup, so changes here seem to have too much effect.

SephReed:
For the changes above, I'd just jumped head first into getting this project to compile:

https://github.com/fuse-box/fuse-mono

It did end up working, but I don't think what I've done is well organized thus far.  I'm currently stepping back and reproaching with a more thorough understanding of resolver.

nchanged:
@SephReed I am going to close it for now. Let's see after we've merged `grand-refactor` into `fuse-compiler` ;_)

PR #1705, opened 2019-12-13
fix(resolver): local modules now find root

SephReed:
All that work, but it ended up being a tiny change.  Also, an insanely difficult bug.

The change is simple, when searching all node_module paths, also check for `node_modules` upwards.


vegarringdal:
sorry never got time.
did you test if the 3 mono repos in the playground works with the edit btw ?

SephReed:
I had not, but because the added code was put directly before an error, anything that didn't error before won't error now.

vegarringdal:
Ok 😁👍
Just wanted to ask since I use 1 common nodemodule folder for all packages/samples.
But Ill try and get time to test it out during the weekend

vegarringdal:
@SephReed 
Give it try and it didnt break any of my weird stuff :-)
Keep the PR comming, great more people are helping @nchanged makign it better :-)

PR #1707, opened 2019-12-16
feature(plugins): Pass plugin options to corresponding transformer

FKobus:
This feature adds the functionality to pass plugin options to the transformer:

Simplified example without fallbacks to defaults:

`fuse.ts`
```js
plugins: [
  pluginTest({
    autoInject: true,
    target: /src\/(.*?)\.(js|jsx|ts|tsx)$/
  })
],
```
`plugin_test.ts`
```ts
export function pluginTest(opts?: any) {
    const { target, ...options } = opts;
    return (ctx: Context) => {
        ctx.transformerAtPath(target, TestTransformer, options);
    };
}
```

`TestTransformer.ts`
```ts
export const TestTransformer = (opts?: any): ITransformer => {
  if (opts.autoInject) {
    // do something else?
  }
}
```

### DIFF
--- a/src/core/Context.ts
+++ b/src/core/Context.ts
@@ -88,9 +88,11 @@ export class Context {
     this.config.setupEnv();
   }
 
-  public transformerAtPath(path: string | RegExp, transformer: (opts: any) => ITransformer) {
+  public transformerAtPath(path: string | RegExp, transformer: (opts: any) => ITransformer, transformerOptions?: any) {
     if (!this._tranformersAtPaths) this._tranformersAtPaths = [];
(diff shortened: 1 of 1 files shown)


SephReed:
This should probably be added to the docs before deleting this branch

https://github.com/fuse-box/fuse-box/blob/master/docs/plugins/pluginCustomTransform.md

nchanged:
@SephReed that's gonna be removed. This is for typescript transformers. There are no docs yet, but I will start working on it soon ;-)

FKobus:
I'll write some documentation for it. I'll use this example.

It's for internal use only atm

PR #1711, opened 2019-12-16
chore: Added doc about transformerAtPath

FKobus:
Based on after merge comments in: https://github.com/fuse-box/fuse-box/pull/1707

### DIFF
--- a/docs/development/transformerPlugin.md
+++ b/docs/development/transformerPlugin.md
@@ -1,0 +1,73 @@
+# Transformer Plugin development
+
+Before starting here, read the part about creating plugins. This document contains a short write up about
+writing a plugin and transformer IN FuseBox. Not for FuseBox. This is only internally accessible.
+
+## Using transformerAtPath
+
+For these plugins we use the internal `transformerAtPath` function. This will assign our transformer thus our 
+plugin to the correct files.
+
+`transformerAtPath` accepts three arguments
+
+```
+public transformerAtPath(
+  path: string | RegExp,
+  transformer: (opts: any) => ITransformer,
+  transformerOptions?: any
+) {};
+```
+
+We use the third parameter to pass down our config.
+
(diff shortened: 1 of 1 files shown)


FKobus:
I think we should create a format for it, because as I wrote, there are multiple ways to invoke it. Maybe good to choose a pattern and enforce that to plugins that use the transformerAtPath

nchanged:
@FKobus I think it depends on what user needs. Maybe a plugin will use a transformer and some internal module operations....It could be both.

SephReed:
From what I know of transformers, I think a hook pattern would be really solid.  I'm hazily imagining something like this:

```
type StandardizedTransformer = Mystery;

interface ITranformerHooks {
  before: {
    all: StandardizedTransformer,
    compile: StandardizedTransformer,
    treeShake: StandardizedTransformer,
    bundler:  StandardizedTransformer,
  },
  after: {
    all: StandardizedTransformer,
    compile: StandardizedTransformer,
    treeShake: StandardizedTransformer,
    bundler:  StandardizedTransformer,
  },
}

transformerThing.addHook({
  before: {
    all: {
     priority: 10,
     fn: (ctx) => doFoo(ctx)
    }
  }
});

transformerThing.addHook({
  before: {
    all: {
     priority: 3.4,
     fn: (ctx) => doBar(ctx)
    }
  }
});
```

This is totally not thought out, more just to show hook pattern.

I've seen this pattern in feathersjs (https://docs.feathersjs.com/guides/basics/hooks.html) and it works really, really well for adding arbitrary transformations to data at arbitrary points.  Very easy to read and understand.

nchanged:
@SephReed It actually makes a lot of sense. I am working on production transformers, and I've realised that I missing those hooks. There are multiple stages. So instead of creating multiple transformers it would make sense to have those handlers.

PR #1712, opened 2019-12-17
Fix: local modules

SephReed:
I found an edge case I'd not covered in my edit.

### DIFF
--- a/src/resolver/nodeModuleLookup.ts
+++ b/src/resolver/nodeModuleLookup.ts
@@ -56,10 +56,10 @@ export function parseAllModulePaths(fileAbsPath: string) {
     if (matchedLast && matchedLast[0]) {
       paths.push(path.join(paths[paths.length - 1], matchedLast[0], 'node_modules'));
     }
-    return paths;
-  } else {
-    return [PROJECT_NODE_MODULES];
   }
+
+  paths.push(PROJECT_NODE_MODULES);
+  return paths;
 }
 
 const CACHED_LOCAL_MODULES: { [key: string]: string | null } = {};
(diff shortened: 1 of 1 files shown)


src/resolver/nodeModuleLookup.ts
```diff
@@ -86,7 +88,7 @@ export function findTargetFolder(props: IResolverProps, parsed: IModuleParsed):
   if (localModuleRoot === undefined) {
     localModuleRoot = CACHED_LOCAL_MODULES[props.filePath] = findUp(props.filePath, "node_modules");
   }
-  if (!!localModuleRoot) {
+  if (!!localModuleRoot && paths.indexOf(localModuleRoot) === -1) {
```
nchanged: is it going to work on windows too? ;-)

SephReed: I think it will actually.  But it'd be better to be explicit about it.

I added this so it wouldn't check a path it already had checked before.  What's the correct way to handle windows paths for this?

vegarringdal: depends on what string values end up in there
path segment is seperated with  \ on Windows and / on LINUX/MAC

vegarringdal: Looked at not node path module, `node.sep` gives you the current on the system..


```js
const path = require('path')
const linux = `/home/project/node_modules`
const window = `\\home\\project\\node_modules`
console.log(linux.split(path.sep).join('/'))  //prints /home/project/node_modules
console.log(window.split(path.sep).join('/')) //prints /home/project/node_modules
```

SephReed: worst case scenario is that it will check the same directory twice.

PR #1713, opened 2019-12-17
Fix: ISDIR: illegal operation on a directory for sha.js

kamilchm:
When package name ends with `.js` like `sha.js` it shouldn't be considered as a module.

### DIFF
--- a/src/resolver/pathsLookup.ts
+++ b/src/resolver/pathsLookup.ts
@@ -1,4 +1,5 @@
 import * as fs from 'fs';
+import * as path from 'path';
 import { fileLookup, ILookupResult } from './fileLookup';
 
 export type ITypescriptPaths = { [key: string]: Array<string> };
@@ -8,6 +9,7 @@ interface IPathsLookupProps {
   isDev?: boolean;
   cachePaths?: boolean;
   paths?: ITypescriptPaths;
+  fromPath: string;
   target: string;
 }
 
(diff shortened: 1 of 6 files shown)


src/resolver/pathsLookup.ts
```diff
@@ -82,6 +84,11 @@ function getIndexFiles(props: IPathsLookupProps): DirectoryListing | undefined {
 }
 
 export function pathsLookup(props: IPathsLookupProps): ILookupResult {
+  const fileDir =
```
nchanged: Are you sure that we need that? Maybe a fix to the regex will do?

kamilchm: IDK, I added a test case based on the structure of `sha.js` to show my motivation for this change. I encounter `ISDIR: illegal operation on a directory for sha.js` when I try to include it in my app bundle.
I didn't find a better way, but everything that will pass the new test case will make me happy.

SephReed:
I'm confused.  Isn't sha.js a module?  (https://github.com/crypto-browserify/sha.js/tree/105bfe57c69e13c83fcf7a6ca660dd984cb291bf)

It has a package.json... and also a file called sha.js.

What it getting to `node_modules/sha.js/sha.js` and treating _that_ as a module?

kamilchm:
In the current version of fusebox, [this](https://github.com/crypto-browserify/sha.js/blob/105bfe57c69e13c83fcf7a6ca660dd984cb291bf/index.js#L10) will be resolved as `node_modules/sha.js` which is a package dir, not a module, so it can't be loaded and the result is an error: `ISDIR: illegal operation on a directory for sha.js`.

nchanged:
@kamilchm can it be reproduced by doing this:
`import "sha.js"` ?

nchanged:
@kamilchm `4.0.0-next.159`

Not reproducible having the following code 

```
import * as oi from 'sha.js';
console.log(oi);

```

kamilchm:
It looks like it's more complicated than I thought before. I encountered it on a large codebase where `sha.js` is deep down in the dependency tree. I'll try to cut only the meaningful parts from there to have the smallest possible sample to reproduce the error.

PR #1715, opened 2019-12-18
Sortier ran on code base

FKobus:
Added sortier to npm dev dependencies
Added sortier run script in package.json

Ran the `npm run sortier` script


FKobus:
never mind.. too many file changes for 1 MR

PR #1734, opened 2020-01-05
feat(web-index): allow HTML template imports

### LINKED ISSUE #1718: [v4] HTML template img tags with src attribute ignored
I have an HTML template that contains a few `img` tags with a relative `src` path set and they are being ignored. I have tried to use the copy plugin to copy those to the dist folder but they are not linked to the output HTML file in any way (i.e. the paths remain the same - no hashes were added).

Is there a way for `fuse-box` to discover those files, add a hash to them in the html file and then copy them to dist?

Forgot to mention, favicons are also ignored.

Example HTML template:
```html
<!DOCTYPE html>
<html>
    <head>
        <title>Example</title>
        <link rel="icon" href="assets/favicon.png" />
        $css
    </head>
    <body>
        <img src="assets/just_an_image.png" />
        <img src="assets/some_graphics.svg" />
        $bundles
    </body>
</html>
```

Teoxoy:
fix #1718

API example:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Example</title>
        <link rel="icon" href="$import('assets/favicon.png')" />
        $css
    </head>
    <body>
        <img src="$import('assets/just_an_image.png')" />
        <img src="$import('assets/some_graphics.svg')" />
        $bundles
    </body>
</html>
```

I reused the `resolveCSSResource` function for this which should be fine for now IMO. A unified way of handling assets/resources would be useful for this though.

I tested this feature in my own project and it ended up working quite nicely.

### DIFF
--- a/src/web-index/webIndex.ts
+++ b/src/web-index/webIndex.ts
@@ -5,6 +5,7 @@ import { env } from '../env';
 import { ensureAbsolutePath, fileExists, readFile, joinFuseBoxPath } from '../utils/utils';
 import { htmlStrings } from './htmlStrings';
 import { FuseBoxLogAdapter } from '../fuse-log/FuseBoxLogAdapter';
+import { resolveCSSResource } from '../stylesheet/cssResolveURL';
 
 export interface IWebIndexConfig {
   enabled?: boolean;
@@ -22,7 +23,7 @@ export interface IWebIndexInterface {
   generate?: (bundles: Array<IBundleWriteResponse>) => void;
 }
 
-export function replaceWebIndexStrings(str: string, keys: { [key: string]: any }) {
+export function replaceWebIndexStrings(str: string, keys: Record<string, string>) {
   return str.replace(/\$([a-z_-]+)/gi, (_var, name) => {
     return keys[name] !== undefined ? (typeof keys[name] === 'object' ? JSON.stringify(keys[name]) : keys[name]) : '';
   });
@@ -101,10 +102,6 @@ export function createWebIndex(ctx: Context): IWebIndexInterface {
           }
         }
       });
-      const scriptOpts: any = {
-        bundles: scriptTags.join('\n'),
-        css: cssTags.join('\n'),
-      };
 
       let fileContents = opts.templateContent;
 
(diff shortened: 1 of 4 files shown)


src/web-index/webIndex.ts
```diff
@@ -120,10 +117,31 @@ export function createWebIndex(ctx: Context): IWebIndexInterface {
...
+        const result = resolveCSSResource(relPath, {
+          contents: '',
+          ctx,
+          filePath: opts.templatePath,
+          options: ctx.config.stylesheet,
+        });
+        if (result) {
+          return result.publicPath;
+        } else {
+          ctx.log.warn(`Unable to resolve ${result.original}`);
```
SephReed: Will this output the full path that was tested?  If not, please add it.  It's a lot easier to tell what when wrong when you know the absolute path that was hit.

Teoxoy: `result.original` does output the absolute path.

SephReed: thank you!  It helps a lot with telling users what they did wrong.

Teoxoy:
For sure :)

Teoxoy:
@nchanged am I right in saying that the hashes are generated based on the absolute path for all resources? I think the hashes of resources should be generated based on modified time or content for best results. Should I open an issue to track this?

nchanged:
The hashes are generated based of the abs path development yes, as for production they should be generated based of their content, but I think it's the same for production now

src/web-index/webIndex.ts
```diff
@@ -120,10 +117,31 @@ export function createWebIndex(ctx: Context): IWebIndexInterface {
         fileContents = pluginResponse.fileContents;
       }
 
+      fileContents = fileContents.replace(/\$import\('(.+?)'\)/g, (_, relPath: string) => {
```
nchanged: One more thing here, we should allow double quotes too.

nchanged: Also, would it make sense to call it "resolve" instead ?

Teoxoy: `resolve` instead of `import`? - I am indifferent

Teoxoy: I didn't want to include double quotes because of this:
image

Teoxoy:
@nchanged I found a problem. When using `runProd` the resources are not copied. Do you know why?

nchanged:
@Teoxoy resources should be copied too.  webIndex is being executed there too. Try tracing it down there must be a switch that wouldn't let it go through.

Teoxoy:
After attaching the web index,

https://github.com/fuse-box/fuse-box/blob/249828fcb82fd5a78888b4df7130de49b0cbb94c/src/production/bundle_prod.ts#L32

`perform` needs to be called and awaited

https://github.com/fuse-box/fuse-box/blob/249828fcb82fd5a78888b4df7130de49b0cbb94c/src/core/ContextTaskManager.ts#L22

---

`perform` is usually called on `complete` and `rebundle_complete`

https://github.com/fuse-box/fuse-box/blob/249828fcb82fd5a78888b4df7130de49b0cbb94c/src/core/ContextTaskManager.ts#L8-L15

but for the production build, `complete` and `rebundle_complete` are called before the web index is attached

@nchanged what would be the best way to fix this?

nchanged:
well the task manager might be copying files, so we don't really need to wait on that. Do you think we can hook into something else? Or shuffle things around?

Teoxoy:
I think we can move this line:

https://github.com/fuse-box/fuse-box/blob/249828fcb82fd5a78888b4df7130de49b0cbb94c/src/production/main.ts#L63

after this one:

https://github.com/fuse-box/fuse-box/blob/249828fcb82fd5a78888b4df7130de49b0cbb94c/src/production/bundle_prod.ts#L32

nchanged:
@Teoxoy sounds good! Move it there and we can merge it

PR #1736, opened 2020-01-05
feat(HMR): add hardReloadScripts option

### LINKED ISSUE #1717: [v4] Hard reload when watched files change
I am developing a WebGL application that can't be easily made to work with HMR. I would like the app to be reloaded for changes to js/ts files and hot reloaded for css files. Is this currently possible?

I have tried `hmrServer: false` option but it ended up not reloading the browser at all.

Related issue for HTML files: #1597

Teoxoy:
fix #1717

### DIFF
--- a/docs/watcher.md
+++ b/docs/watcher.md
@@ -45,15 +45,3 @@ fusebox({
 ```
 
 *(note: The `ignored` field given to `chokidar` is declared by the `watch` field in FuseBox's configuration.  Do not define the `ignored` field here.)*
-
-### Hard reload scripts (js/ts files)
-
-By default scripts are hot reloaded. Change the `hardReloadScripts` option to hard reload when changes are detected to script source files.
-
-```ts
-fusebox({
-  watch: {
-    hardReloadScripts: true,
-  },
-});
-```
--- a/docs/watcher.md
+++ b/docs/watcher.md
@@ -45,3 +45,15 @@ fusebox({
 ```
 
 *(note: The `ignored` field given to `chokidar` is declared by the `watch` field in FuseBox's configuration.  Do not define the `ignored` field here.)*
+
+### Hard reload scripts (js/ts files)
+
+By default scripts are hot reloaded. Change the `hardReloadScripts` option to hard reload when changes are detected to script source files.
+
+```ts
+fusebox({
+  watch: {
+    hardReloadScripts: true,
+  },
+});
+```
--- a/src/hmr/attach_hmr.ts
+++ b/src/hmr/attach_hmr.ts
@@ -2,8 +2,11 @@ import { Context } from '../core/Context';
 import { Module } from '../core/Module';
 import { Package } from '../core/Package';
 import { fastHash, measureTime } from '../utils/utils';
+import { EXECUTABLE_EXTENSIONS } from '../config/extensions';
 import { WatcherAction } from '../watcher/watcher';
 import { generateHMRContent } from './hmr_content';
+
+const SCRIPT_EXT_REGEX = new RegExp(`\\.(${EXECUTABLE_EXTENSIONS.join('|').replace(/\./g, '')})$`);
 
 function generateUpdateId() {
   return fastHash(new Date().getTime().toString() + Math.random().toString());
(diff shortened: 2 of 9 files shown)


Teoxoy:
Locally, I managed to hard reload when the HTML template changes too but the problem is that even though it's being reloaded, it has the old content - it's not being read again from the fs

src/hmr/attach_hmr.ts
```diff
@@ -108,10 +108,20 @@ export function attachHMR(ctx: Context) {
...
+    devServer.clientSend('reload', undefined, ws_instance);
+  }
+
   // here we recieve an update from client - the entire tree of its modules
   devServer.onClientMessage((event, payload: IClientSummary, ws_instance?: WebSocket) => {
-    if (event === 'summary' && payload.id && tasks[payload.id]) {
-      softProjectUpdate(tasks[payload.id], payload, ws_instance);
+    const task = tasks[payload.id];
+    if (event === 'summary' && payload.id && task) {
+      if (task.action === WatcherAction.HARD_RELOAD_MODULES) {
```
nchanged: Hm, no that's not going to work

src/watcher/watcher.ts
```diff
@@ -54,19 +58,25 @@ export enum WatcherAction {
   RESTART_PROCESS = 'RESTART_PROCESS',
 }
 
-export function detectAction(file: string, homeDir: string): WatcherAction {
-  const isInsideHomeDir = !/\.\./.test(path.relative(homeDir, file));
-
+export function detectAction(file: string, homeDir: string, hardReloadScripts: boolean): WatcherAction {
```
nchanged: I would leave this unchanged

src/hmr/attach_hmr.ts
nchanged: You need to send an event like this:
https://github.com/fuse-box/fuse-box/blob/master/src/hmr/attach_hmr.ts#L144

Then pick it up here:

https://github.com/fuse-box/fuse-box/blob/master/modules/fuse-box-hot-reload/index.js#L21

The best would be to undo the changes in this block too.

nchanged:
```
client.on('reload', () => window.location.reload())
```
something like that ;-)

Teoxoy: I added that in the last commit. I got confused what hard reloading ment in the context of the watcher. Should the check for hard reloading go in the hmr module or is it fine how it is?

nchanged: well, hard reloading means ignoring the cache in this context. I would undo the changes and made a condition where I pointed out. So instead of sending a request for summary just send an event to reload Done ;-) 5 lines of code ;-)

Teoxoy: Oh ok

Teoxoy:
Where should I add the docs for this? I only added the option to the full-config markdown file. There is no separate file for the hmr module.

docs/watcher.md
```diff
@@ -44,4 +44,4 @@ fusebox({
 });
 ```
 
-*(note: The `ignored` field given to `chokidar` is declared by the `watch` field in FuseBox's configuration.  Do not define the `ignored` field here.)*
+*(note: The `ignored` field given to `chokidar` is declared by the `watch` field in FuseBox's configuration.  Do not define the `ignored` field here.)*
```
nchanged: I think you removed the docs ;-)

Teoxoy: I removed the docs from the watcher because I moved the config option to the hmr prop in the config.
Should I make a new doc file for the hmr options?

nchanged: sure, would be nice to the docs in place!

nchanged: You can make a new file `hmr.md` whatever works best. We can refactor it later.

Teoxoy: What does `reloadEntryOnStylesheet` do? - I want to include it in the doc

nchanged: hey, join slack https://slack.fuse-box.org/ it's easier to coordinate there. It's when you'd want to reload an entry point when a css file is changed

PR #1738, opened 2020-01-06
fix(production): polyfill global

### LINKED ISSUE #1654: global not  polyfilled in browser runProd
Some packages that uses global (global.localStoreage etc) works in runDev  but global is undefined in runProd

Teoxoy:
fix #1654

Uses `globalThis` for modern browsers and [this polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#Examples) for those that don't have `globalThis`.

### DIFF
--- a/src/production/transformation/processTransformation.ts
+++ b/src/production/transformation/processTransformation.ts
@@ -21,7 +21,7 @@ function onGlobal(node: Identifier, props: LocalContext) {
     }
     if (!props.globalInserted) {
       props.globalInserted = true;
-      props.insertions.push(`const global = {};`);
+      props.insertions.push(`const global = globalThis || self || window || global;`);
     }
   }
 }
--- a/src/production/transformation/processTransformation.ts
+++ b/src/production/transformation/processTransformation.ts
@@ -21,7 +21,12 @@ function onGlobal(node: Identifier, props: LocalContext) {
     }
     if (!props.globalInserted) {
       props.globalInserted = true;
-      props.insertions.push(`const global = globalThis || self || window || global;`);
+      props.insertions.push(`const global =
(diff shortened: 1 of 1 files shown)


src/production/transformation/processTransformation.ts
```diff
@@ -21,7 +21,7 @@ function onGlobal(node: Identifier, props: LocalContext) {
     }
     if (!props.globalInserted) {
       props.globalInserted = true;
-      props.insertions.push(`const global = {};`);
+      props.insertions.push(`const global = globalThis || self || window || global;`);
```
nchanged: you actually need to check them for being undefined. Otherwise, it will throw an error
```
typeof window !== "undefined" ? window : ...
```

Teoxoy: Why will it throw an error?

This works:
```js
let a = undefined
let b = undefined
let c = { prop: 6 }
let d = { prop2: 12 }
let g = a || b || c || d
// g = c
```

nchanged: `let g = a || b || c || d`

this will throw ^ Since those variables are not defined

nchanged: You can  open a debugger a type `global` you will see what I mean. Either way, I would just assign global to window if defined. If I am not mistaken that is being applied only for browser builds?

Teoxoy: Oh yeah, I don't know what I was thinking.

Teoxoy: Regarding your question, NodeJS has `global` and in workers you can't use `window` only `self`. So I think the order is good:
`globalThis` else `self` else `window` else `global`

PR #1739, opened 2020-01-06
Suggested RegEx patterns to ignore

paxperscientiam:
Flycheck will at times create temporary files. This might have the undesired effect of triggering FuseBox watch mechanism. To avoid this, the pattern `/flycheck_/i` could be used.


nchanged:
@paxperscientiam do you think you can add this RegEX to the default ignored paths?

paxperscientiam:
@nchanged , I could for sure.

paxperscientiam:
will submit a PR

PR #1741, opened 2020-01-06
Feat: production warmup imports

FKobus:
# Imports transformer for production
- create moduleTree
- cover most common ways of imports
- wrote tests to validate code is correct

Run only the tests in this PR with: 
`npx jest src/production/transformers/__tests__/Phase_1_Import.test.ts`


src/production/module/ImportReference.ts
```diff
@@ -0,0 +1,268 @@
...
+
+
+function ImportSpecifier(visit: IVisit, specifier: ASTNode): ImportSpecifierReference {
+  let local: string;
+  let name: string;
+
+  if (specifier.type === 'ImportNamespaceSpecifier') {
+    // import * as React from 'react'
+    local = specifier.local.name;
+    name = 'default';
```
nchanged: oops, `name` shouldn't even be here. This will lead to errors, We should instead set `importNamespace : true` IMHO.
but name should be definitely `undefined`
Suggestions
`importNamespace` `importAll`

src/production/module/ImportReference.ts
```diff
@@ -0,0 +1,268 @@
...
+export interface ImportSpecifierReferenceProps {
+  local: string;
+  name: string;
+  specifier: ASTNode;
+  visit: IVisit;
+};
+
+export function ImportSpecifierReference(props: ImportSpecifierReferenceProps) {
+  return {
+    local: props.local,
```
nchanged: `importAll`  boolean flag here?

src/production/transformers/__tests__/Phase_1_Import.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    const { tree } = test(`
+      import * as bar from './foo';
+    `);
+    const refs = tree.importReferences.references;
+
+    expect(refs).toHaveLength(1);
+    expect(refs[0].source === './foo');
+    expect(refs[0].specifiers).toHaveLength(1);
+    expect(refs[0].specifiers[0].local === 'bar');
+    expect(refs[0].specifiers[0].name === 'default');
```
nchanged: should be `undefined`

src/production/module/ImportReference.ts
FKobus: added types for specifiers:

```ts
export enum ImportSpecifierReferenceType {
  OBJECT_SPECIFIER,
  NAMESPACE_SPECIFIER
}
```

src/production/module/ImportReference.ts
FKobus: see previous comment, types added

src/production/transformers/__tests__/Phase_1_Import.test.ts
FKobus: source code updated. name is now undefined

PR #1747, opened 2020-01-08
chore: refactor return type imports

FKobus:


src/production/module/ImportReference.ts
```diff
@@ -222,11 +239,14 @@ function exportSpecifierImport(props: IImportReferencesProps, scope: IImportRefe
...
-  const references: Array<IImport> = [];
+export interface IImportReferences {
+  references: Array<IImport>;
+  register: (props: IImportReferencesProps) => void;
+}
 
+export function ImportReferences(productionContext: IProductionContext, module: Module): IImportReferences {
   const scope = {
-    references,
+    references: [],
```
FKobus: not sure if i need to default it to an empty array

nchanged: why not? xD

PR #1749, opened 2020-01-09
feat: CodeSplittingPhase

FKobus:
This PR is the base for code splitting on production.

Todo:
- ~~add more tests~~
- ~~prevent circular dependencies~~


src/production/__tests__/CodeSplitting.test.ts
```diff
@@ -0,0 +1,347 @@
...
+        import './b';
+      `,
+    });
+    expect(context.modules).toHaveLength(8);
+    // we only have 1 split
+    expect(context.splitEntries.entries).toHaveLength(1);
+    // the split we have has 7 modules: a, b, c, d, e, f, g
+    expect(context.splitEntries.entries[0].modules).toHaveLength(7);
+  });
+});
```
nchanged: I would add a few cases with `require` and `export {foo} from "./foo"` just in case ;-)

FKobus: Those are already covered in the Import tests, but i'll add them for 'just in case' ;)

FKobus: ✅ added extra test with different import / require types

src/compiler/transformers/astHelpers.ts
```diff
@@ -1,9 +1,12 @@
-import { ASTNode } from '../interfaces/AST';
 import * as path from 'path';
+import { ASTNode } from '../interfaces/AST';
+
 export function getDynamicImport(node: ASTNode) {
-  // buntis case
+  // meriyah case
   if (node.type === 'ImportExpression') {
-    if (node.source) return { source: node.source.value };
+    if (node.source) {
```
FKobus: @nchanged  somewhere this changed, do you prefer the singleline if?

src/production/docs/codesplitting.md
```diff
@@ -0,0 +1,61 @@
...
+}
+```
+
+```js
+splitEntries: {
+  entries: [{
+    entry: { // FooModule:
+      moduleTree: {
+        ...,
+        moduleType: SPLIT_MODULE
```
FKobus: seems overkill if a module is in the splitEntries... thinking about removing the moduleType here..
Or do you need it in a later process? @nchanged

src/production/module/SplitEntries.ts
```diff
@@ -0,0 +1,50 @@
...
+  return {
+    entry: module,
+    modules: subModules,
+    references: module.moduleTree.dependants,
+  };
+}
+
+export interface ISplitEntries {
+  entries: Array<ISplitEntry>;
+  ids: { [key: number]: boolean };
```
FKobus: not sure if this is a good name for an object of moduleIds... It's used for quick lookup if a module is a splitEntry

src/production/phases/CodeSplittingPhase.ts
```diff
@@ -0,0 +1,189 @@
...
+    },
+    subModules,
+    /**
+     * traceCircularDependency accepts a target of type Module
+     * it needs a reference to the parentId so we can safely trace back
+     *
+     * @param target
+     * @param parentId
+     */
+    traceCircularDependency: function (target: Module, parentId: number): boolean {
```
FKobus: I can merge this function into traceOrigin...

Downside, traceOrgin will look more complicated and harder to understand in the future...

@nchanged any preference?

src/production/phases/CodeSplittingPhase.ts
```diff
@@ -0,0 +1,189 @@
...
+     * and returns true if it origins back to the possibleSplitEntry Module
+     *
+     * @param target
+     */
+    traceOrigin: function (target: Module, parentId: number): boolean {
+      const { moduleId, moduleTree: { dependants } } = target;
+      // console.log(`traceOrigin ${moduleId} ${target.getShortPath()}`);
+
+      if (moduleId in this.visited) {
+        // @todo: log circular trace to console
```
FKobus: should i implement a logging for this? where the trace is outputted for the user?

`a.js -> b.js -> c.js -> a.js`

src/production/phases/CodeSplittingPhase.ts
```diff
@@ -0,0 +1,189 @@
...
+          // the module is a dynamic module and not the entry, so false!
+          // we flagAsCommonsEligible here!
+          target.flagAsCommonsEligible();
+          traced = false;
+        } else {
+          traced = this.traceOrigin(module, moduleId);
+        }
+        if (!traced) {
+          // @todo: config to flag all shared commons?
+          // target.flagAsCommonsEligible();
```
FKobus: i kinda want this 😇

src/production/phases/CodeSplittingPhase.ts
```diff
@@ -0,0 +1,189 @@
...
+     * to see if we can integrate those too
+     *
+     * @param target
+     */
+    traverseDependencies: function (target: Module, entry: boolean = false): boolean {
+      const { moduleId, moduleTree: { importReferences: { references } } } = target;
+
+      // we already traversed this module, so we can skip it
+      // this happens in case of circular deps
+      if (!entry && moduleId in this.traversed) {
```
FKobus: `'key' in 'object'` vs `!!object[key]`

src/production/docs/codesplitting.md
nchanged: I think it's needed yes

src/production/module/SplitEntries.ts
nchanged: ids is fine. Just use `Record<number, boolean>` looks better ;-)

src/production/phases/CodeSplittingPhase.ts
nchanged: I like the way it is now ;-)

src/production/phases/CodeSplittingPhase.ts
nchanged: we could technically pass a logger object... to log those

src/production/phases/CodeSplittingPhase.ts
nchanged: `object[key]` unless your key can be `0`

FKobus: key can be 0, because we start from 0... i can make the moduleId counter start at 1?

src/production/__tests__/CodeSplitting.test.ts
```diff
@@ -0,0 +1,444 @@
...
+          const dynamicFunc = await import('./bar');
+        }
+
+        console.log(test);
+      `,
+      'foo.ts': `
+        export function foo() {}
+      `,
+      'bar.ts': `
+        import { test } from './test';
```
nchanged: This module isn't in use

FKobus: yes it is, in index.ts => `console.log(test)`

FKobus: true, extended the test!

src/production/docs/codesplitting.md
FKobus: Leaving it alone then ;)

src/production/module/SplitEntries.ts
FKobus: ✅ done

src/production/phases/CodeSplittingPhase.ts
FKobus: ✅ me too

PR #1766, opened 2020-01-14
feat: computed statement support

FKobus:
Work in progress


src/production/__tests__/ComputedStatementImport.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    expect(result).toEqual(['./atoms/', expressionValueTypes.DOUBLE_ASTERISK, '/', expressionValueTypes.SINGLE_ASTERISK]);
+  });
+
+  it('import(`./atoms/${b + \'/\' + c}`) should return a valid pattern', () => {
+    // Tests the case:
+    // import(`./atoms/${b + '/' + c}`);
+    const result = convertComputedStatementToRegex(getAst("`./atoms/${b + '/' + c}`"));
+    expect(result).toEqual(['./atoms/', expressionValueTypes.DOUBLE_ASTERISK, '/', expressionValueTypes.SINGLE_ASTERISK]);
+  });
+});
```
nchanged: I would add tests for non-supported nodes, like function calls and such.

FKobus: consider it done

src/compiler/__tests__/computed.statement.import.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    // import('./assets/' + var + '.svg');
+    const result = computedStatementToPath(getAst(`'./assets/' + a + '.svg'`));
+    expect(result).toEqual(['./assets/', expressionValueTypes.DOUBLE_ASTERISK, '.svg']);
+  });
+
+  it(`import('./' + path + (() => 'test')); should return an error object`, () => {
+    // Tests the case:
+    // import('./' + path + (() => 'test'));
+    const result = computedStatementToPath(getAst(`'./' + path + (() => 'test')`));
+    expect(result).toEqual({ error: `Unsupported type provided to computed statement import` });
```
nchanged: There is one thing I don't like about it. Is that it can return both an array or an error. The interface should be 
```
{
  error?: string;
  paths ?: Array<string>  
}
```

FKobus: updated the code base. I was annoyed by it too

PR #1767, opened 2020-01-15
fix(compiler): speed improvements

nchanged:


src/compiler/transpilers/testTranspiler.ts
```diff
@@ -27,7 +28,12 @@ export function testTranspile(props: ICompileModuleProps) {
 
   const ext = props.fileName ? path.extname(props.fileName) : '.tsx';
 
-  const ast = parseTypeScript(contents, { jsx: ext !== '.ts' });
+  let ast;
+  if (props.useMeriyah) {
+    ast = parseJavascript(contents, { jsx: ext !== '.ts' });
```
FKobus: shouldn't this be `.js`

nchanged: should be always true ;-) js can contain jsx ;-)

PR #1774, opened 2020-01-21
feat: refactor production test and improvs

FKobus:
TODO:

- splitted modules as own bundle
- multiple entries
- commons


src/production/ProductionContext.ts
```diff
@@ -1,30 +1,51 @@
...
+}
+
+export function createProductionContext(ctx: Context, modules?: Array<IModule>): IProductionContext {
+  let entries;
+
+  /**
+   * @todo:
+   * Not sure about this check
+   */
+  if (!modules) {
```
FKobus: from tests we provide it with a modules array...
do you have an idea how to solve this more properly?

nchanged: we need to throw an exception... Probably in ModuleResolver, a fatal error that the files weren't resolved

src/production/phases/FinalPhase.ts
```diff
@@ -1,20 +1,43 @@
...
+  await ctx.ict.resolve();
+
+  for (const module of modules) {
     if (module.isExecutable) {
-      module.transpile();
+      const transformerResult = module.transpile();
+
+      for (const item of transformerResult.requireStatementCollection) {
+        item.statement.callee.name = BUNDLE_RUNTIME_NAMES.ARG_REQUIRE_FUNCTION;
+        item.statement.arguments[0].value = module.moduleSourceRefs[item.statement.arguments[0].value].id;
```
nchanged: so many things can go wrong here ;-)

FKobus: i know.. will fix this (added a comment)

src/core/fusebox.ts
```diff
@@ -1,57 +1,54 @@
...
         props.handler(new UserHandler(ctx));
       }
-      bundleProd(ctx);
 
-      // return bundleProd(ctx);
-      return Promise.resolve();
+      try {
+        await bundleProd(ctx);
+      } catch (e) {
+        ctx.log.error(`Fatal error: ${e}`);
```
nchanged: `ctx.fatal`

src/moduleResolver/ModuleResolver.ts
```diff
@@ -80,22 +80,32 @@ function initModule(props: { absPath: string; bundleContext: IBundleContext; ctx
...
-          statement: source,
-        });
-
-        module.moduleSourceRefs[source] = resolvedModule;
-        // re-writing the reference
-        item.statement.callee.name = BUNDLE_RUNTIME_NAMES.ARG_REQUIRE_FUNCTION;
-        item.statement.arguments[0].value = resolvedModule.id;
-        delete item.statement.arguments[0].raw;
+        // no value.. import() or require();
+        if (!item.statement.arguments[0]) continue;
```
nchanged: we still need to make an `ARG_REQUIRE_FUNCTION` of it. And pass `-1` or `0` as an argument. (IMHO) so we could pick it up and runtime and throw an error

src/moduleResolver/ModuleResolver.ts
```diff
@@ -80,22 +80,32 @@ function initModule(props: { absPath: string; bundleContext: IBundleContext; ctx
...
+            bundleContext,
+            ctx,
+            importType: item.importType,
+            parent: module,
+            statement: source,
+          });
+          if (!resolvedModule) continue;
+          module.moduleSourceRefs[source] = resolvedModule;
+          // re-writing the reference
+          item.statement.callee.name = BUNDLE_RUNTIME_NAMES.ARG_REQUIRE_FUNCTION;
```
nchanged: How about making this re-usable?

src/moduleResolver/ModuleResolver.ts
```diff
@@ -80,22 +80,32 @@ function initModule(props: { absPath: string; bundleContext: IBundleContext; ctx
...
+          // we're importing some weird stuff... require(1) ie.
+          if (typeof source !== 'string') continue;
+          const resolvedModule = resolve({
+            bundleContext,
+            ctx,
+            importType: item.importType,
+            parent: module,
+            statement: source,
+          });
+          if (!resolvedModule) continue;
```
nchanged: Let's make sure that if we're not able to resolve that module we do:

* What I mentioned above
*  Also make sure that logging is in place

PR #1782, opened 2020-01-29
feat: module naming

FKobus:


nchanged:
@FKobus the tests have failed. Easy to fix, but please fix the snapshots manually to avoid some unpredicted updates

PR #1784, opened 2020-01-30
feat: server entry

FKobus:


src/production/module/ServerEntry.ts
```diff
@@ -0,0 +1,25 @@
+import { BundleType, createBundle, IBundleWriteResponse } from '../../bundle/bundle';
+import { Context } from '../../core/context';
+
+export async function createServerEntry(ctx: Context, bundles: Array<IBundleWriteResponse>) {
+  const serverEntry = createBundle({
+    bundleConfig: {
+      path: '_serverEntry.js',
```
nchanged: let's make in configurable in IRunProps `bundles`

PR #1787, opened 2020-02-01
Refactor/improved cache

nchanged:


src/cache/cache.ts
```diff
@@ -31,190 +35,334 @@ export interface ICacheMeta {
   packages: Record<string, IPackage>;
 }
 
-export function getMTime(absPath): number {
-  return statSync(absPath).mtime.getTime();
+function readJSONFile(target: string) {
```
FKobus: shouldn't this be a function in utils?

src/testUtils/integrationTest.ts
```diff
@@ -1,91 +1,194 @@
...
+        },
+      );
+      var fn = new Function('window', '__fuse', app);
+      const errors = [];
+      try {
+        fn(window, __fuse);
+      } catch (e) {
+        errors.push(e);
+      }
+      return new Promise((resolve, reject) => {
```
FKobus: you can omit the return and reject. 

`return new Promise(resolve => resolve({...})`

nchanged: Actually cleaner to have it in array, and jsom had a handler for that

PR #1788, opened 2020-02-02
[v4] WIP: Support for Yarn v2

mekwall:
This PR add initial support for Yarn v2 Plug'n'play module resolution.

Note that `typescript` needs to be a dependency and not only a devDependency. This is because `@typescript-eslint/typescript-estree` requires `typescript` as a peerDependency and Yarn v2 will need fuse-box to provide it.


src/resolver/nodeModuleLookup.ts
```diff
@@ -123,7 +134,7 @@ export function nodeModuleLookup(props: IResolverProps, parsed: IModuleParsed):
   if (!fileExists(packageJSONFile)) {
     return { error: `Failed to find package.json in ${folder} when resolving module ${parsed.name}` };
   }
-  const json = require(packageJSONFile);
```
nchanged: I am afraid that just passes and absolute path which won't affect the require logic

mekwall: Ah ok, then it probably isn't necessary. I'll try without it!

PR #1811, opened 2020-02-22
fix: dev server should send 404 code for not existed static files

starsolaris:


src/devServer/devServer.ts
```diff
@@ -57,7 +57,7 @@ export function createExpressApp(ctx: Context, props: IHTTPServerProps, extra?:
   app.use('/', express.static(props.root));
 
   app.use('*', (req, res) => {
-    res.sendFile(props.fallback);
+    res.status(404).sendFile(props.fallback);
```
nchanged: I don't think thats correct, since all single page apps resolve their route paths to index

starsolaris: But this is should be not default configuration and should be configured by user manually.
When i make GET http://localhost:4444/some.png i should receive image or error but not html page without 200 ok.

nchanged: fair enough, let me think about it. We'd also need to spit out a warning in this case.

PR #1813, opened 2020-02-24
feat: css in jsx

FKobus:


src/compiler/transformers/shared/CSSinJSXTransformer.ts
```diff
@@ -0,0 +1,245 @@
...
+              if (autoLabel && !parent.callee && isEmotionCall(node, 'callee')) {
+                labelMapping['[local]'] =
+                  (parent.type === 'VariableDeclarator' && parent.id.name) ||
+                  (parent.type === 'AssignmentExpression' && parent.left.property.name) ||
+                  '';
+                node.arguments.push(renderAutoLabel());
+              }
+              break;
+
+            case 'JSXElement':
```
nchanged: what's the purpose of this block?

src/compiler/transformers/shared/CSSinJSXTransformer.ts
```diff
@@ -0,0 +1,245 @@
...
+ * 1. expand the minify (way to simple :P)
+ *
+ * 2. Components as selectors
+ * 3. Minification
+ * 4. Sourcemaps
+ * 5. Minify configurable?
+ *
+ * 5. Dead Code Elimination // not needed??
+ */
+export function CSSInJSXTransformer(options?: CSSInJSXTransformerOptions): ITransformer {
```
nchanged: also please create integration tests (similar to plugin angular)

src/compiler/transformers/shared/CSSinJSXTransformer.ts
FKobus: Ah, that's one of the features not implemented yet.. it's called css prop optimization.

Commented out, but will be used when i start implementing missing features

PR #1814, opened 2020-02-24
feat: update dependencies

patrickmichalina:


nchanged:
@patrickmichalina some typings have changed, the devServer is broken with these updates

Issue #1793, opened 2020-02-06
v4 feature request: support lerna packages and yarn workspaces


martaver:
So the new approach to using typescript's paths property for monorepo works nicely.

The truth is, though, that for a real world monorepo setup, I wouldn't use typescript paths at all. My monorepos are set up with lerna and yarn workspaces, which symlinks the source code from child package to a parent's node_modules directory. The parent references the child package in their package.json as 'just another dependency'.

This allows each dependency to be treated as an independent package and has many benefits, including being able to automatically calculate change logs and semver and navigate dependency trees. In fact, yarn workspaces already keeps the dependency tree for us.

The downside to this, is that to fuse-box, the child packages look like 'just another dependency', and fuse-box can't tell the difference between our monorepo packages and those installed in node_modules through npm.

The result of this is that fuse-box builds everything correctly, but we get no watch or hmr on our local monorepo dependencies.

Ideally, we would update usages of ´resolver.ts´ to allow for yarn workspace and lerna workspace information to be read, so that we can detect when dependencies are local, and treat them the same as though they were included in tsconfig paths.

Martaver:
To clarify, if I combine yarn workspaces with paths in tsconfig, everything seems to work fine.

mekwall:
Is this with yarn v1 or v2? Afaik paths in tsconfig doesn't work with Yarn v2.

Martaver:
Hmm, I'm still on yarn 1... I'm not familiar with yarn 2 and how it might change the monorepo game.

Basically, I think yarn workspaces are the correct way to do deps between packages in monorepos. It opens up lots of cool stuff like automatic dependency walking when generating semver from commits etc.

The thing is that fuse-box won't listen to changes inside packages linked via yarn workspaces because they just look like any other node_module. So I have to add paths in tsconfig as well to get fuse-box to work.

The drawback then, is that I no longer actually need to add the referred package as a dependency to my code. I want to maintain that need for explicit declaration of deps, so that semver can be calculated, and if we need to split up the monorepo, any package can be pulled out and replaced with npm or verdaccio on a whim.

Martaver:
To achieve this, basically all fuse-box would have to do is look up the workspaces properties for lerna/yarn from package.json or lerna.json, then iterate the package names from each of them - and it'd have a list of packages that it should watch when encountered. Seems like small potatoes. I'd take a look myself if I could get some guidance on where to start.

nchanged:
[...]

PR #1820, opened 2020-02-29
fix: use node module resolution

warrenfalk:
(as discussed on Slack)

Module resolve in fuse-box does not match node's module resolution.  (This is one of the key issues with #1793 which mentions workspaces.  Workspaces and Lerna rely on node's module resolution).

For example, if you were to require `ImportMe` from the file `/home/alovelace/fuseapp/file.js` then node would search the following paths:

1. `/home/alovelace/fuseapp/node_modules/ImportMe`
2. `/home/alovelace/node_modules/ImportMe`
3. `/home/node_modules/ImportMe`
4. `/node_modules/ImportMe`

But fusebox will currently not search above its own root.  So it only searches the first path.  This can be worked around somewhat by changing the root, but this impacts other things and has its own cons.

This PR modifies the test case to match node's behavior and patches the behavior to pass the test.  

Also as requested on slack, for performance, a wrapper function was added which checks the existence of these paths, remembers the result, and only returns the paths that exist.

### DIFF
--- a/src/resolver/__tests__/nodeModuleLookup.test.ts
+++ b/src/resolver/__tests__/nodeModuleLookup.test.ts
@@ -4,6 +4,7 @@ import '../../utils/test_utils';
 import { findTargetFolder, isNodeModule, parseAllModulePaths, nodeModuleLookup } from '../nodeModuleLookup';
 import { ensureDir } from '../../utils/utils';
 import { createRealNodeModule } from '../../utils/test_utils';
+import { resolve as pathResolve } from 'path'
 
 const cases = path.join(__dirname, 'cases');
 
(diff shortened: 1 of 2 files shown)


src/resolver/nodeModuleLookup.ts
```diff
@@ -36,29 +36,32 @@ export function isNodeModule(path: string): undefined | IModuleParsed {
...
+  if (exists === undefined) {
+    exists = fileExists(absPath);
+    pathExists.set(absPath, exists);
   }
+  return exists;
+}
 
-  return [PROJECT_NODE_MODULES];
+export function parseExistingModulePaths(fileAbsPath: string): string[] {
+  return parseAllModulePaths(fileAbsPath).filter(memoizedExists);
```
nchanged: That looks legit, except `filter` is very slow. If you noticed we're trying to avoid using those function in favour of for loops;-)
https://jsperf.com/for-vs-filter

That's a nitpick tho. If you can fix it that would be great ;-)

src/resolver/nodeModuleLookup.ts
```diff
@@ -36,29 +36,32 @@ export function isNodeModule(path: string): undefined | IModuleParsed {
   return result;
 }
 
-export function parseAllModulePaths(fileAbsPath: string) {
+export function parseAllModulePaths(fileAbsPath: string): string[] {
   const baseDir = path.dirname(fileAbsPath);
+  const segments = baseDir.split(path.sep).filter(s => s);
```
nchanged: nit: same here. Filters add overhead

src/resolver/__tests__/nodeModuleLookup.test.ts
```diff
@@ -114,20 +114,29 @@ describe('NodeModule lookup', () => {
...
-    expect(paths[3]).toMatchFilePath(`a/node_modules/@angular/core/node_modules/foo/node_modules/bar/node_modules$`);
-  });
-
-  it('Should give a primary path ', () => {
-    const path = 'some/user/file';
+    expect(paths).toStrictEqual([
+      '/node_modules',
+      '/a/node_modules',
+      '/a/node_modules/@angular/node_modules',
+      '/a/node_modules/@angular/core/node_modules',
```
nchanged: I think these tests will fail on windows, that's why there is a helper `toMatchFilePath `

warrenfalk: I don't know why I didn't think to check Windows.

Yes, but in fact, it's not just the test that is wrong.  The function itself will not work right on Windows because of the different behavior of absolute paths, which on windows contains a drive name.  I did just push a version that works on windows, but discovered it doesn't work in posix.  I'm going to redo the whole function and then I'll resolve this.

PR #1841, opened 2020-03-10
feat: always check for changes in directories  

starsolaris:
add possibility for checking dirs for changes without cache

### DIFF
--- a/src/cache/cache.ts
+++ b/src/cache/cache.ts
@@ -11,6 +11,7 @@ export interface ICachePublicProps {
   FTL?: boolean;
   enabled: boolean;
   root?: string;
+  alwaysCheck?: string[];
 }
 
 export interface IModuleRestoreResponse {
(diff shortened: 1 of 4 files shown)


src/cache/cache.ts
```diff
@@ -11,6 +11,7 @@ export interface ICachePublicProps {
   FTL?: boolean;
   enabled: boolean;
   root?: string;
+  alwaysCheck?: string[];
```
nchanged: Let's rename it to `safeCheckModules`

src/cache/cache.ts
```diff
@@ -291,7 +292,8 @@ export function createCache(ctx: Context, bundleContext: IBundleContext): ICache
       const target = modules[depId];
       if (!target) return;
       const pkg = packages[target.packageId];
-      if (pkg && pkg.isExternalPackage) {
+      const alwaysCheck = pkg && ctx.config.cache.alwaysCheck.length > 0 && ctx.config.cache.alwaysCheck.filter((name) => pkg.publicName.includes(name)).length > 0;
```
nchanged: let's not use `filter` but a `for of` loop instead.

starsolaris:
Ok, i will create new pull request to master.

PR #1858, opened 2020-03-17
fix: do not fall back to node_modules in a PnP environment, and don't hide PnP errors

warrenfalk:
In a PnP environment you don't want to fall back on node_modules search logic.  If PnP says that a dependency can't be resolved, then it shouldn't be resolved.  PnP is designed to really own module resolution.

PnP is intentionally stricter about loading dependencies.  For example, it won't allow a module in one package to import another package unless the first one has declared the second as a dependency.  Being able to "find" that dependency doesn't matter.  If the environment is PnP and PnP says it can't be resolved, then it shouldn't be resolved.


nchanged:
But how will we resolve internal modules then ? There has to be a workaround with an exception here

warrenfalk:
Internal modules are resolved before checking PnP so they still override PnP.

PR #1859, opened 2020-03-17
feat: allow specifying watcher root paths

warrenfalk:
This is a proposed feature to how watcher is configured.

Chokidar lets you specify one to many "root" paths to watch.  Currently fuse-box watches only one root, which is always `env.APP_ROOT`.  This PR leaves that as the default but allows overriding it with a `root` property in `IWatcherPublicConfig`.

Note that in a monorepo environment relying on workspaces, you would specify those workspace folders as roots, and in a future change, I plan to add the ability to auto-detect if in a workspaces environment and add those automatically.  But manual override of roots is a good feature to have anyway.

### DIFF
--- a/docs/watcher.md
+++ b/docs/watcher.md
@@ -14,12 +14,25 @@ fusebox({ watcher: true });
 By default the watcher watches the entire application root, but reacts only to the path relative to the first entry
 point.
 
-You can change that by passing `include` property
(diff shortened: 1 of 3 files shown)


nchanged:
@warrenfalk could you update the docs for this? I don't want to have a documentation debt ;-)

warrenfalk:
Yep, doc updated

PR #1860, opened 2020-03-17
feat: use typescript tsconfig refs

warrenfalk:
This pull request has three things in it (because these three things cannot be easily be separated).  It might be easier to review the commits independently.

  1. **fix: return real paths from nodeModuleLookup**
      * This just calls `realpathSync()` on the results of `nodeModuleLookup()`.  This will have no effect unless the path contains a symlink, in which case it dereferences the symlinks and returns the "real" path.  This is what node's `resolve()` and `require()` do.  So this is more node-resolve-compatibility  This behavior is essential for non-PnP monorepo setups.
  2. **chore: simplify logic in nodeModuleLookup**
      * This cleans up the code in `nodeModuleLookup()` to remove a bunch of duplicated code and logic.  This change should not have any behavioral effect, it just simplifies the code.
      * One function signature was also modified to narrow an argument from an object down to the one field that the function used.  A test was updated.
  3. **feat: use typescript tsconfig references**
      * This is the main change, more about this below.

These changes are essential in supporting of monorepos as mentioned in #1793

**TypeScript Project References**

This change allows fuse-box to read the `"references"` field of a `tsconfig.json`.  This change should have no effect on existing fuse-box users unless they are using this field (which is very unlikely if they are using fuse-box, since fuse-box doesn't support the field today).

The main change is a new module, `typescriptReferences.ts` that can recurse through the `"references"` from `tsconfig.json` to `tsconfig.json` which results in building a map of output files (like `dist/file.js`) to the input file that builds them (like `src/file.ts`) even when the outputs don't exist.  This map is then used to construct a SubPathResolver that will look for input files when someone imports an output file.

Tests were added to check `typescriptReferences.ts`.

`fileLookup()` has been modified to take a `SubPathResolver` which it uses to check the existence of files in a package.  If none is specified, it defaults to using a `SubPathResolver` called `resolveIfExists` which just does what `fileLookup()` was doing before.  When using tsconfig references, the typescript subpath resolver gets passed in.

All existing tests pass unmodified.


nchanged:
@warrenfalk I have a question here. You are using real typescript API to resolve it, right?. What's is the performance overhead? For example, if one isn't using `references` it won't kick in, right?

src/resolver/nodeModuleLookup.ts
```diff
@@ -112,104 +112,82 @@ export function findTargetFolder(props: IResolverProps, parsed: IModuleParsed):
...
-    }
-
-    result.isEntry = false;
-    result.targetFuseBoxPath = makeFuseBoxPath(folder, result.targetAbsPath);
-  } else {
-    const entryFile = getFolderEntryPointFromPackageJSON({ isBrowserBuild: isBrowser, json: json });
 
-    const entryLookup = fileLookup({ fileDir: folder, target: entryFile });
+  const isEntry = !target;
+  const targetResolver = props.tsTargetResolver || fileLookup;
```
nchanged: so if references are set it's going to use typescript for resolving the files? Could it potentially disrupt the lookup defined by FUseBox, like aliases e.t.c

src/compilerOptions/typescriptReferences.ts
```diff
@@ -0,0 +1,222 @@
...
+                if (result) {
+                    return result;
+                }
+            }
+        }
+        return undefined;
+    }
+
+    // Construct a TargetResolver but inject our own TypeScript sub-path resolver
+    return (lookupArgs: ILookupProps) => {
```
nchanged: Just to be clear this tsResolver is a wrapper for `fileLookup` ?!

src/resolver/nodeModuleLookup.ts
warrenfalk: Aliases are processed first, and the output of processing aliases is what is sent to `nodeModuleLookup()`.  So aliasing can interfere with Typescript, but not the other way around.

The TypeScript resolver just does its own aliasing and passes the result onto the default resolver.  Its own aliasing is automatically aliases output files to input files.  So the order here means that if you alias to an output file, it will be aliased again (by typescript) to an input file.  I believe this is the desired behavior.

For some examples:
  1. If I alias `"file.js"` to `"dist/file.js"` then the typescript resolver will further change it from `"dist/file.js"` to `"src/file.ts"` and load that.
  2. If I alias `"file.ts"` to `"src/file.ts"` then the typescript resolver will just pass that through unchanged because it isn't an output.

The only potential issue, I think, is if someone tries to alias an input file to another input file, but still tries to import the output file. But I don't think this is a problem, because it already wouldn't work today, and it still is the expected behavior, I think.  The scenario looks like this:
  1. If I alias `"src/file.ts"` to `"src/other-file.ts"` and then I `import "src/file.ts"` this will work, but if I `import "dist/file.js"` then typescript will change it to `"src/file.ts"` but it won't go *back through* aliasing, so what will be loaded is `"src/file.ts"` not `"src/other-file.ts"`.

nchanged: Sounds awesome man, one thing I'd ask you to do to help me review it.
Could you add a comprehensive example into the `playground` folder? I will check out the branch and see the work in action.

You're becoming the man behind the fuse-box module resolution ;-) I trust your skills, but I want a playground example mainly because I want to understand how it works!

src/compilerOptions/typescriptReferences.ts
warrenfalk: Yes, but it injects its own `subPathResolver` instead of the default.

The reason this is done this way is that we want to use all of the logic inside fileLookup (e.g. look for exact match, then with extension, then index.js, etc.) but we want every one of those lookups to see if there is an associated input file and load that instead.

So for example, if someone `import 'mypkg'` then this will call into `fileLookup` and will read the `package.json` which has a `"main"` of `"dist/index.js"` and when it checks the subpath resolver to see if that exists, typescript can look for `"src/index.ts"` instead.

warrenfalk:
> @warrenfalk I have a question here. You are using real typescript API to resolve it, right?. What's is the performance overhead? For example, if one isn't using `references` it won't kick in, right?

Correct.  We use the real typescript API once at the beginning when processing the main `tsconfig.conf` file just for building the mappings, and after that we just look everything up in the already-built mappings.

Correct, this code is only entered if someone has `"references"` in their `tsconfig.json`.

I figure it shouldn't affect anyone because probably no one that uses fuse-box is using `"references"`.  And when people do start to use it, if performance is an issue, we can look at perhaps re-implementing and optimizing that logic.

src/resolver/nodeModuleLookup.ts
warrenfalk: Yeah, I should be able to do that.  I have my own playground project I have been working with, so I will polish that up a bit and see if it will do the trick.

src/compilerOptions/typescriptReferences.ts
warrenfalk: Playground examples added.

There's `playground/ts-monorepo-1` which is Yarn 1, and `playground/ts-monorepo-2` which is Yarn 2.

Note that the Yarn 2 playground package might not work correctly due to the placement of the `playground` folder in the repo.  Yarn 2 can find the `package.json` of `fuse-box` which will cause it to complain.

warrenfalk: There is still one more problem to resolve here which can be demonstrated in the playground example.

If you run `playground/ts-monorepo-1` and make a change to `@myorg/bar/src/watchtrigger.ts`, you can see the watcher pick it up (which is good) but the change doesn't make it through cache/hmr.

This seems to be because `@myorg/bar` is not seen as a "user" package.  I have a (proposed) fix for this, which I'll submit in a separate PR for discussion.

nchanged: Awesome! I will check it tomorrow on Monday! There are files that do some simulations? Could you leave me instructions on what to run first?

nchanged:
@wagerfield please rebase against master. And could you work with fuse-box repository directly? You have the permissions to push to branches except for master.

src/compilerOptions/typescriptReferences.ts
warrenfalk: As I was writing some instructions, it occurred to me I could make some things easier so I am modifying the playground code to be easier to use.  Both playgrounds are basically the same, but `-2` is tweaked for yarn 2.  Both of them have READMEs in them that have some explanation and instructions.  I'm going to make some quick improvements to these, and update this PR in the next few minutes.

Also: I can't easily submit a separate PR for my proposed fix for the "user / non-user package issue" because it is based too heavily on changes in this PR and can't be separated.  So I'm going to put it back in this PR.  My change works and all tests pass, but I want to make sure it is heading in the right direction.

warrenfalk:
Rebased.

I can work out of the fuse-box repo from now on.  Thanks, I hadn't realized I had that ability.

I pushed a `feature/tsconfig-refs` to the fuse-box repo to see if I could modify this PR to use it, but it looks like github doesn't support changing the source branch for a PR.  Do you want me to close this one and open another?

I also included `feat: add detection of user/non-user package dependencies` on here for discussion.  This is what I referred to elsewhere.  I'm looking for feedback on the approach taken in that commit.

src/moduleResolver/asyncModuleResolver.ts
```diff
@@ -66,8 +66,11 @@ export async function asyncModuleResolver(ctx: Context, entryFiles: Array<string
       }
       let pkg = bundleContext.getPackage(resolved.package.meta);
 
+      const isUserPackage = resolved.package.isUser;
+      const pkgType = isUserPackage ? PackageType.USER_PACKAGE : PackageType.EXTERNAL_PACKAGE;
+
       if (!pkg) {
-        pkg = createPackage({ meta: resolved.package.meta, type: PackageType.EXTERNAL_PACKAGE });
+        pkg = createPackage({ meta: resolved.package.meta, type: pkgType });
```
warrenfalk: This is the main change for the user/non-user detection.  We set the type to `USER_PACKAGE` based on whether `isUser` was set during package resolution.

Here I am understanding a package to be a "User" package if it is expected that the user will modify the package.  So this should be true for custom modules workspace packages, but false for registry-downloaded packages.

src/resolver/nodeModuleLookup.ts
```diff
@@ -72,14 +72,22 @@ export function parseExistingModulePaths(fileAbsPath: string): string[] {
   return existing;
 }
 
+function isDirectory(path: string) {
+  return fileExists(path) && lstatSync(path).isDirectory();
+}
+
 const CACHED_LOCAL_MODULES: { [key: string]: string | null } = {};
-export function findTargetFolder(props: IResolverProps, name: string): string {
+export function findTargetFolder(props: IResolverProps, name: string): { error: string } | { folder: string, isUser: boolean } {
```
warrenfalk: This function is where the detection happens, so I modified the return type to include `isUser: boolean`.

Also I made it so that when it returns an error, it returns the error message.

Note: I also modified the return type so that it returns *either* an error or a real result, so that TypeScript warns if you try to access the `folder` prop if you haven't first checked the `error` prop.

src/cache/cache.ts
```diff
@@ -280,6 +280,20 @@ export function createCache(ctx: Context, bundleContext: IBundleContext): ICache
         }
       }
     }
+
+    // for any dependencies that have package.json
+    // if the package.json changes, then the "main" might change
+    // which might change the target.absPath
+    // so we can no longer trust the dependencies and have to look them up again
```
warrenfalk: This is the change I was least certain about.  There is code elsewhere that decides whether to call `restorePackage()` or to call `restoreModuleSafely()`.  I mostly left that code how it is, but then one of the tests fails when it modifies a `package.json` because of the issue described in this comment.  So I added this code here which resolves it, but I'm not sure whether this is the best place or not.

src/compilerOptions/typescriptReferences.ts
nchanged: @wagerfield I am going to start reviewing in in a few hours. If you need to continue with other changes, it's probably better if you base off of this branch. It is, however, a fork. That's why I was suggesting moving it to a fuse-box repo itself.

nchanged:
@warrenfalk I followed your instructions
Screenshot 2020-03-23 12 34 54

But this what I see.

nchanged:
Also, please use `prettier` `sortier` and `editorconfig` extensions for VSCode to keep it consistent. 
Your editor settings vary drastically, those extensions would format everything according to `.editorconfig`

src/resolver/nodeModuleLookup.ts
```diff
@@ -117,104 +125,93 @@ export function findTargetFolder(props: IResolverProps, parsed: IModuleParsed):
...
-  return result;
+  const targetExtension = path.extname(targetAbsPath);
+
+  return {
+    isEntry,
+    targetAbsPath,
+    targetFuseBoxPath,
+    targetExtension,
+    meta: pkg,
+    isUser,
```
nchanged: shall we give it a better name? I took a while for me to figure it out

nchanged:
@wagerfield there is a noticeable performance perf. degradation. compare angular_playground which doesn't use references. From 3s 300ms to 3s 500ms 

That's a tiny difference but will add up. The culprit here is the number of calls which has increased by a good amount. 

We should try decreasing the number of function calls, as well as get rid of `map` `reduce`  e.t.c those are the performance killers.

src/resolver/nodeModuleLookup.ts
```diff
@@ -117,104 +125,93 @@ export function findTargetFolder(props: IResolverProps, parsed: IModuleParsed):
...
-    }
-
-    result.isEntry = false;
-    result.targetFuseBoxPath = makeFuseBoxPath(folder, result.targetAbsPath);
-  } else {
-    const entryFile = getFolderEntryPointFromPackageJSON({ isBrowserBuild: isBrowser, json: json });
-
-    const entryLookup = fileLookup({ fileDir: folder, target: entryFile });
+  const isEntry = !target;
+  const targetResolver = (isUser && props.tsTargetResolver) || fileLookup;
```
nchanged:
```
let targetResolver = fileLookup;
if( isUser && props.tsTargetResolver ) targetResolver = props.tsTargetResolver
```

This is much faster. Also `||` is slower comparing to a regular ternary

src/resolver/resolver.ts
```diff
@@ -155,8 +156,8 @@ export function resolveModule(props: IResolverProps): IResolver {
       }
       const pkg = nodeModuleLookup(props, moduleParsed);
 
-      if (pkg.error) {
-        return { error: pkg.error };
+      if ("error" in pkg) {
```
nchanged: any particular reason that was changed?

nchanged:
@warrenfalk let's chat on Slack, we would need a point of reference when it gets to performance. Some sort of test.  Here is a summary:

* Please fix your VSCode formatting installing the tools I mentioned above
* Minor changes that improve the performance
* Let's think about reducing the number of function calls in places which a called many times throughout the resolver lifecycle
* And finally IMHO some variables need to be renamed. But if you have a strong opinion on those, let's have a chat on Slack

src/core/context.ts
```diff
@@ -145,6 +151,9 @@ export function createContext(props: ICreateContextProps): Context {
 
   self.compilerOptions = createCompilerOptions(self);
 
+  // calculate typescript output->input mappings from references
+  self.tsTargetResolver = self.compilerOptions.tsReferences && createTsTargetResolver(self.compilerOptions.tsReferences, self.compilerOptions.tsConfig || path.join(env.SCRIPT_FILE));
```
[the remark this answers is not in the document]
warrenfalk: Maybe `tsSubPathResolver`...

No, this isn't really related to node modules.  This is based on the `rootDir` and `outDir` of `tsconfig.json`; It uses the `references` to find all the `tsconfig.json` files.

So if it sees:
```json
"rootDir": "src",
"outDir": "dist",
```

It builds a resolver that translates the sub-path (the "target") below:
```ts
import { thingy } from "mypkg/dist/mymodule"
//                            ^^^^^^^^^^^^^-- only deals with this part
```

The `"mypkg"` part is the node module part, and it has already been resolved before this resolver gets called.

This resolver just uses what it learned from `rootDir` and `outDir` to change the `dist/mymodule` part to `src/mymodule.ts`.  _Notice that it goes from `/dist/` to `/src/` and from a file that would have a `.js` extension to a `.ts` file._

I'm still open to suggestions on the name.

src/resolver/nodeModuleLookup.ts
warrenfalk: Maybe `isUserModule`?

The point is that this module is one of the user's own modules.  I.e. one that the user will edit.  And so it should be treated differently during caching when scanning for changes.

But I'm certainly open to suggestions on this one too.  But I'm going to submit this one later in a separate PR anyway

src/resolver/resolver.ts
warrenfalk: To satisfy TypeScript's type checker.

Here's a quick explanation

image

warrenfalk:
Closing this in lieu of #1874, which also has a tweak to angular playground so you can test performance on that.

PR #1868, opened 2020-03-22
chore: cleanup remove unreachable code and yarn-error.log

warrenfalk:
While looking at other things, I found some code that is unreachable.  This PR just removes the code since it isn't reachable.  But you'll want to review to see whether the code was just accidentally left it, or if it should be moved elsewhere instead of removed.

I also removed a yarn-error.log which I don't think was committed intentionally.


warrenfalk:
After I typed the description I realized I should also add yarn-error.log to the .gitignore, so I did that, too.

src/cache/cache.ts
```diff
@@ -220,9 +220,9 @@ export function createCache(ctx: Context, bundleContext: IBundleContext): ICache
       const depPackage = packages[meta.packageId];
       // a required dependency is missing
       // verifying and external package of the current package
-      if (!meta) return false;
```
warrenfalk: As you can see a few lines above, `meta.packageId`. Which will throw if `meta` is falsey.  So this check here doesn't do anything.  The `return false` is unreachable.

src/cache/cache.ts
```diff
@@ -246,11 +246,6 @@ export function createCache(ctx: Context, bundleContext: IBundleContext): ICache
     const meta = findModuleMeta(absPath);
 
     const metaPackage = packages[meta.packageId];
-    // not present in meta
-    if (!meta) {
```
warrenfalk: `meta.pacakgeId` above will have thrown already if `!meta`, so this code block isn't reachable.

PR #1872, opened 2020-03-22
Package manager suggestions

paxperscientiam:


package.json
```diff
@@ -138,5 +138,5 @@
     "typecheck": "ts-node -T fuse typecheck"
   },
   "typings": "index.d.ts",
-  "version": "4.0.0-next.415"
-}
\ No newline at end of file
+  "version": "4.0.0-alpha.329"
```
nchanged: this is not correct

paxperscientiam: fixed!

PR #1879, opened 2020-03-26
fix: allow providing relative root for watcher

warrenfalk:
Currently relative root paths don't work in the watcher config.  E.g. the following doesn't work

```javascript
watcher: {
    root: "..",
}
```

This PR adds support for relative paths.

### DIFF
--- a/src/watcher/watcher.ts
+++ b/src/watcher/watcher.ts
@@ -2,7 +2,7 @@ import { watch as chokidarWatch } from 'chokidar';
 import * as path from 'path';
 import { Context } from '../core/context';
 import { env } from '../env';
-import { ensureScriptRoot, path2RegexPattern } from '../utils/utils';
+import { ensureScriptRoot, path2RegexPattern, excludeRedundantFolders } from '../utils/utils';
 import { bindWatcherReactions } from './bindWatcherReactions';
 export type IWatcher = ReturnType<typeof createWatcher>;
 
@@ -43,9 +43,15 @@ export function createWatcher(ctx: Context) {
   let includePaths: Array<RegExp> = [];
   let ignorePaths: Array<RegExp> = [];
 
+  const { root } = props;
+  // ensure root is string[]
+  const roots = typeof root === "string" ? [root] : root;
+  // ensure roots are absolute paths
+  const absRoots = roots && roots.map(r => ensureScriptRoot(r));
+
   if (!props.include) {
-    if (props.root) {
-      includePaths = (typeof props.root === "string" ? [props.root] : props.root).map(path2RegexPattern)
+    if (absRoots) {
+      includePaths = absRoots.map(path2RegexPattern)
     }
     else {
       // taking an assumption that the watch directory should be next to the entry point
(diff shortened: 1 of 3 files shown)


src/watcher/watcher.ts
```diff
@@ -43,9 +43,14 @@ export function createWatcher(ctx: Context) {
   let includePaths: Array<RegExp> = [];
   let ignorePaths: Array<RegExp> = [];
 
+  // ensure root is string[]
+  const root = typeof props.root === "string" ? [props.root] : props.root
+  // ensure roots are absolute paths
+  const absRoots = root && root.map(r => path.resolve(`${env.SCRIPT_PATH}`, r));
```
nchanged: let's use `ensureScriptRoot` 
this will allow users to use abs paths

src/watcher/watcher.ts
```diff
@@ -43,9 +43,14 @@ export function createWatcher(ctx: Context) {
   let includePaths: Array<RegExp> = [];
   let ignorePaths: Array<RegExp> = [];
 
+  // ensure root is string[]
+  const root = typeof props.root === "string" ? [props.root] : props.root
```
nchanged: [].concat(props.root)

src/watcher/watcher.ts
```diff
@@ -146,8 +151,8 @@ export function createWatcher(ctx: Context) {
       const userOptions = props.chokidarOptions || {};
       const finalOptions = { ...defaultOpts, ...userOptions };
 
-      const root = props.root || [`${env.APP_ROOT}`];
-      const watcher = chokidarWatch(root, finalOptions);
+      const watchRoot = absRoots || `${env.APP_ROOT}`;
```
nchanged: shouldn't absRoots be defined before this line? The default config should be
`env.SCRIPT_PATH`

src/watcher/watcher.ts
warrenfalk: Ah, cool, didn't know that was there, I'll change that.

By the way, a cool feature of `path.resolve()` is that it already handles abs paths... 

image

src/watcher/watcher.ts
warrenfalk: I avoided `[].concat(x)` because the type of that expression is `any[]`

image

image

warrenfalk: Oh, also... more importantly `props.root` might be `undefined`

src/watcher/watcher.ts
warrenfalk: I didn't change the default, it was already `env.APP_ROOT` so I left it.  I can change it if you think I should, but when I thought about it, `APP_ROOT` made sense as it is probably more likely to do what the user wants since in almost all cases, `SCRIPT_PATH` is inside `APP_ROOT` anyway, and if the user wants fine-grained control, they can mess with `include`/`ignore`

`absRoot` can be `undefined` here because `root` might be `undefined` because `props.root` might be `undefined`.

And so if the user doesn't specify a root, this is where the defaulting ends up happening.

nchanged: sorry, I mean `watchRoot` should always contain APP_ROOT regardless of user settings. The watcher detects changes to `fuse.ts`. So we should always have it there.

warrenfalk: Gotcha, I'll get that

warrenfalk: This became a little more complicated.

Because `SCRIPT_PATH` might be the same as `APP_ROOT` and so you can easily end up with:

```
[
  `/my/app/path`,
  `/my/app/path`, // <-- redundant
]
```

But removing duplicates isn't really enough because one root might contain the other.  I.e. `SCRIPT_PATH` might be _inside_ `APP_ROOT`

```
[
  `/my/app/path`,
  `/my/app/path/scripts` // <-- also redundant
]
```
So the best solution is to remove all _redundant_ paths.  I added a utility function, `excludeRedundantFolders()` to do that and a test to check that it works.

PR #1897, opened 2020-04-20
fix: stylus dependency filename change

tcrowe:
## What does it do?

The stylus people changed from `stylus/lib/nodes/Literal.js` to `literal.js`. So one character changed. This PR updates fuse-box stylus renderer to use that file.

Before this PR:

`❌ Cannot find module 'stylus/lib/nodes/Literal'`

After:

Screen Shot 2020-04-19 at 7 47 52 PM

Screen Shot 2020-04-19 at 7 49 28 PM


tcrowe:
Related stylus project file: https://github.com/stylus/stylus/blob/dev/lib/nodes/literal.js

Issue #1834, opened 2020-03-08
[v4] .sass files treated as .scss


bsl-zcs:
```
body
  color: red
```

`❌ Invalid CSS after "b": expected 1 selector or at-rule, was "body" in ...`

tried to pass `indentedSyntax: true` to `pluginSass` - it doesn't help

gleba:
i have same message with **FuseBox 4.0.0-next.413**

please fix support **.sass** files

gleba:
![image](https://user-images.githubusercontent.com/1052840/80565019-1a0a5180-89f8-11ea-991a-ddb96d5babfa.png)
indentedSyntax absolutely unsupported now

gleba:
option indentedSyntax is really missed in node-sass render function

gleba:
fixed in **^4.0.0-next.430**
@bsl-zcs @charsleysa 
now u can resolve/close issue

PR #1901, opened 2020-04-29
Fix: SASS indent syntax support

gleba:
fix for issue  **.sass files treated as .scss**  (https://github.com/fuse-box/fuse-box/issues/1834)

### DIFF
--- a/src/stylesheet/sassHandler.ts
+++ b/src/stylesheet/sassHandler.ts
@@ -80,6 +80,7 @@ export async function renderModule(props: IRenderModuleProps): Promise<IStyleshe
     includePaths: [path.dirname(module.absPath)],
     outFile: module.absPath,
     sourceMapContents: requireSourceMap,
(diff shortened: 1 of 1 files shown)


gleba:
@nchanged pls review

src/stylesheet/sassHandler.ts
```diff
@@ -80,6 +80,7 @@ export async function renderModule(props: IRenderModuleProps): Promise<IStyleshe
     includePaths: [path.dirname(module.absPath)],
     outFile: module.absPath,
     sourceMapContents: requireSourceMap,
+    indentedSyntax: module.absPath.split('.').pop() == "sass",
```
nchanged, line 83: how about `module.extension === ".sass"` ?

gleba: Yes, it's right!
I made changes according to the git flow.
just one commit in logs

PR #1914, opened 2020-05-19
Fix/project refs target server

Martaver:
Previously, when target=server, fuse-box would ignore all external dependencies by default `serverIgnoreExternals`:`true`. This is because in general, we don't want to bundle node_modules dependencies.

When using project references, however, this means that our referenced projects are also ignored, since they are considered external.

In the special case that they are actually 'user owned' projects, fuse-box should actually build and bundle them.

This PR implements this special case, so that even when `serverIgnoreExternals` is `true`, project references are bundled anyway.


src/testUtils/integrationTest.ts
```diff
@@ -131,7 +133,7 @@ export function createTestWorkspace(props: {
   let cacheFolder = path.join(rootDir, '.cache');
 
   // creating modules if needed
-  const modulesRoot = path.join(rootDir, 'modules');
+  const modulesRoot = path.join(rootDir, MODULES_PATH);
```
nchanged, line 136: I think those were `modules` on purpose ;-) you changed those to node_modules

package.json
```diff
@@ -140,5 +140,5 @@
     "typecheck": "ts-node -T fuse typecheck"
   },
   "typings": "index.d.ts",
-  "version": "4.0.0-next.433"
+  "version": "4.0.0"
```
nchanged, line 143: please revert that;-)

src/testUtils/integrationTest.ts
nchanged: @Martaver any reason you changed that to node_modules? Could you revert it?

PR #1952, opened 2020-10-02
fix: apply macros before checking for special words

starsolaris:
Apply macros before checking for svg filters, http links, data:image (base64), sass and less variables.
Without this check url cannot be corrected with macros in some cases.


nchanged (approved):
👍

PR #1953, opened 2020-10-08
code splitting broken for electron production build

Zaitsev:


patrickmichalina (commented):
LGTM

nchanged (approved):
👍

vegarringdal:
Ive always just used  target "server" for backend/preloads and "browser" for frontend when using electron.

Zaitsev:
By docs its 'electron' target so I believe it should work 
BTW I saw some node packages optimisations in fuse for electron (ingnorePackages or something, [electronStore](https://github.com/sindresorhus/electron-store) doesn't works with 'browser', some node integrations )

vegarringdal:
ok :-)
Keep helping making fusebox better 👍 

Btw, have not used electronStore.
Basicly check if its frontend or preload/backend and use single bundle and ignoreAllexternal for all
Also only use native modules in preload/backend.
```js
 useSingleBundle: isFrontEnd ? false : true,
 dependencies: { ignoreAllExternal: true },
```

Guess my config is kinda messy 😂 
use backend, frontend/preload and own preload for expressjs for calling database etc so I dont lock frontend/backend

weird little app I made that helps me at work...
image

PR #1956, opened 2020-10-13
fix: circular reference detection along branch

### LINKED ISSUE #1954: Incorrect circular reference detection
**Fuse-Box version**: 4.0.0-next.444

I have a composite project with a folder structure as follows:

```
core/
  |- common
  |- logger
  |- handshakes
services/
  | ...
web/
```

Both `web` and `services` have dependencies on packages in the `core` folder (hence separation), but inside `core` packages can still have one-directional dependencies on other packages inside `core`. In particular, the webpage dependency graph looks like this:

```
   / -> core/common/
  /         ^
web/ -> core/logger/
  \         ^
   \ -> core/handshakes/
```

I can run a build on the project from the web folder and TSC will compile everything with no complaints of circular references; however, when I try to run a fuse-box build it spits out an error because the webpage has a direct dependency on `core/common`, and a direct dependency on `core/logger` which *also* has a dependency on `core/common`.

```
Error: Project references may not form a circular graph. Cycle detected: core\common\tsconfig.json
```

Obviously, the first thing I did was double check my dependencies to make sure there wasn't actually a circular reference, because I have had TSC happily compile circular references before and then everything blows up at runtime.

Once I'd verified that all the dependency graphs were one-way and still got the error, I dug into the fuse-box files and worked around this issue by letting it dig all the way down through a branch in a dependency, checking for circular references within that branch as it goes, instead of checking for a reference anywhere else in the tree as it is currently doing.

The changes I ended up making in fuse-box to make my project work are as below:

```ts
// src/compilerOptions/typescriptReferences.ts#buildMappings
export function buildMappings(references: ITsConfigReference[], tsConfigDir: string): ByOutputMap {
  // build the mappings
  const tsHost = createTsParseConfigHost();
  const inputsByOutput = new Map<string, string>();
  const rawReferences = references || [];
  for (const rawRef of rawReferences) {
    const { path: rawPath } = rawRef;
    if (!rawPath) continue;
    const absPath = ensureAbsolutePath(rawPath, tsConfigDir);
    recurseTsReference(tsHost, absPath, inputsByOutput, [realpathSync(absPath)]);
  }
  return inputsByOutput;
}

// src/compilerOptions/typescriptReferences.ts#recurseTsReference
function recurseTsReference(tsHost: ParseConfigHost, reference: string, map: ByOutputMap, anticycle: string[]) {
  if (!fileExists(reference)) {
    throw new Error(`Unable to find tsconfig reference ${reference}`);
  }
  const realPath = realpathSync(reference);
  const result = loadTsConfig(realPath, tsHost);
  if (!result) return;
  const { files, path, references } = result;
  for (const file of files) {
    const existing = map.get(file.output);
    // Ensure either there is no entry for that output, or if there is, it is the same input producing it
    if (existing && existing !== file.input) {
      throw new Error(
        `Multiple input files map to same output file (1. "${existing}", 2. "${file.input}") => ("${file.output}")`,
      );
    }
    map.set(file.output, file.input);
  }
  const referenceAnticycle = [];
  for (const subref of references || []) {
    const subrefAnticycle = recurseTsReference(tsHost, subref.path, map, [...anticycle, reference]);
    for (const ref of subrefAnticycle) {
      if (anticycle.includes(ref)) {
        throw new Error(`Project references may not form a circular graph. Cycle detected: ${ref}`);
      }
    }
    referenceAnticycle.push(realpathSync(subref.path));
  }
  return referenceAnticycle;
}
```

RevanProdigalKnight:
Changes code to detect circular references along a given branch instead of looking for other references to the same project in the entire tree.

Fixes #1954

### DIFF
--- a/src/compilerOptions/typescriptReferences.ts
+++ b/src/compilerOptions/typescriptReferences.ts
@@ -37,13 +37,12 @@ export function buildMappings(references: ITsConfigReference[], tsConfigDir: str
   // build the mappings
   const tsHost = createTsParseConfigHost();
   const inputsByOutput = new Map<string, string>();
-  const anticycle = new Set<string>();
   const rawReferences = references || [];
   for (const rawRef of rawReferences) {
     const { path: rawPath } = rawRef;
     if (!rawPath) continue;
     const absPath = ensureAbsolutePath(rawPath, tsConfigDir);
-    recurseTsReference(tsHost, absPath, inputsByOutput, anticycle);
+    recurseTsReference(tsHost, absPath, inputsByOutput, [realpathSync(absPath)]);
   }
   return inputsByOutput;
 }
@@ -142,7 +141,7 @@ function packageAncestorsOf(path: string): { base: string; rel: string }[] {
 
 // Follow a single TypeScript reference recursively through its references
 // adding output->input mappings along the way
-function recurseTsReference(tsHost: ParseConfigHost, reference: string, map: ByOutputMap, anticycle: Set<string>) {
+function recurseTsReference(tsHost: ParseConfigHost, reference: string, map: ByOutputMap, anticycle: string[]) {
   if (!fileExists(reference)) {
     throw new Error(`Unable to find tsconfig reference ${reference}`);
   }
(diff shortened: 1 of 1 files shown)


nchanged (approved):
👍 LGTM

PR #1962, opened 2020-11-20
fix wrong test of element exist in array

Kend0r:
hello
splitFileNames is an array so splitFileNames[fileName] will always respond with undefined, use indexOf instead.

My problem was only visible when build on linux, on windows, the relativePath come with "/" as separator, or 'path.sep' is "\" so paths.shift() respond with full path, or in linux 'path.sep' is "/" so paths.shift() respond with "/" and the split works, and becaue of wrong test in the while loop, the function return always the fileName only without path, so that make same name for all my dynamic package (because I remove the hash tag too)
thx


nchanged (approved):
👍

PR #1969, opened 2020-12-12
V4docs added to website folder

vegarringdal:
I really want v.4.0.0 tag on fusebox before Christmas, so we get out of the never ending next we are in.

* updated dependencies under website folder (still docusaurus v1)
* added docs folder root files to v4.0.0 in website
* made plugins part of main docs..
* deleted old_docs folder at root
* deleted files under docs folder at root, since Ive added them to 4.0.0 and added ids etc
  * what is the docs folder used for? temp storage, or can this override docs?

Im sure the docs needs more work, but with this we have at least something for release..
I suck at making docs, and have very little time, but hopefully this will make it easier for others to help.

how to run for everyone else
* go to website folder, run `npm run start:dev`

image

image


nchanged (approved):
Thanks a lot! That looks good to me! ;-)

charsleysa:
This PR is outside of my knowledge area for this project so I'll refrain from fully reviewing it.

There is one aspect that I think should change though. If you need to keep an empty folder and make sure it's committed to the repo you should create an empty file in the folder named `.gitkeep`.

vegarringdal:
.gitkeep sounds nicer.
was hoping @nchanged knew why we needed the folder.
first I though it was the mail doc folder.
and maybe it is and I just have messed up the sidebar.js file.
never used the doc lib before. Pretty much just want us out of alpha/next cycle. And for that we need main website updated with some docs (they need more work, I just copied what we had)

vegarringdal:
@nchanged 
Renamed to .gitkeep.
Suggest you take a hour with quick fixes and update website.
I would just release fusebox to version 4 now so everyone have a better version to relate to.
Patch on a stable version is more fun to work with. Because its VERY stable for my use.

Future of fusebox with new tech like esbuild/snowpack is something that needs to be addressed too.
If v4 is the final version, I think its important to make it clear within next 1-3 months.

Fusebox have been amazing, but web is forever changing so there is no doubt esbuild is a game changer.

PR #1975, opened 2021-01-15
fix: upgrade all dependencies

patrickmichalina:
`"meriyah": "4.0.0"` is the big one that should fix an issue I am having.


vegarringdal (approved):
👍 
I dont have any problems with this.

charsleysa:
@patrickmichalina the version restrictions mean that patch versions cannot be used, would it not be better to use `^4.0.0` instead? or even `~4.0.0`? Same for chokidar

patrickmichalina:
Yeah good point I was just copying the previous format. Can change it though

nchanged:
@patrickmichalina is it good to merge?

patrickmichalina:
@nchanged yeah. I updated the resolutions as recommended by @charsleysa

PR #1988, opened 2021-03-02
fix: optional chaining

### LINKED ISSUE #1981: Optional chaining emits undefined when used with a function call
Using code like the following:
```
const text = useField('query')?.input?.value;
```
Results in:
```
const text = (_2_ = (_1_ = undefined('query')) === null || _1_ === void 0 ? void 0 : _1_.input) === null 
```
See that `useField` is being emitted as `undefined` and then called.

charsleysa:
Fixes #1981


nchanged (approved):
Good job!

nchanged:
Thanks a lot for doing it! it's good to go!

PR #1998, opened 2021-05-14
Correct config.hmr type

paxperscientiam:
So, this is a quick fix. However, it seems there are quite a few inconsistences in docs.

Here's another if anyone's able to address it:

On docs/development/full-config ...
Screen Shot 2021-05-14 at 2 55 50 PM
... the hmr link points to 

https://fuse-box.org/docs/development/hmr
Screen Shot 2021-05-14 at 2 56 43 PM
However, 1) there is no such property `config.watch` according to the `IPublicConfig`, and 2) there's nothing explicitly about HMR on this page.


nchanged (approved):
👍

PR #1999, opened 2021-05-27
fix: before_webindex_write applied for consolidate

zeniroy:
this make pluginConsolidate works(like ejs)


nchanged (approved):
👍 looks good!

PR #2001, opened 2021-05-27
fix: empty reactionStack bug

zeniroy:
sometimes reaction is fired after reactionStack is empty. so clone stack for reaction.


nchanged (approved):
👍

PR #2002, opened 2021-05-27
fix: var scope calculation

zeniroy:
declaration with var keyword has function scope. additional process needed


nchanged:
@zeniroy amazing job! But can you cover it with tests? It's hard to follow what's the issue, so the tests would actually help!

nchanged:
@zeniroy `src/compiler/__tests__/scopes.test.ts`

zeniroy:
ok i will add some test soon.

nchanged (approved):
👍

PR #2003, opened 2021-05-27
fix: wait for first bundle completed

zeniroy:
sometimes additional process needed after bundle completed. so when build first bundle, wait all the events ends.


nchanged (approved):
👍 looks good!

PR #2004, opened 2021-05-27
fix: when no cache, module id changed on rebuild

zeniroy:
when no cache, module id changed on rebuild.
so i added another cache for module id cacheing.


nchanged:
hm.. what's the impact of that on the performance?

zeniroy:
i think performance is not affected. because id cache is so simple and made just once on first build time.

nchanged (approved):
👍

PR #2005, opened 2021-05-27
fix: css module order bug

zeniroy:
css module resolving order is so important.
so i made module id same with resolving order and when building css, reordering css modules by module id.


nchanged:
It's looking good! And I think we could put that to the test. Could you take a look at it?
I suppose we could add them to:

```
src/production/__tests__/CodeSplitting.test.ts
```

zeniroy:
ok i will add some test soon.

nchanged (approved):
👍  GOOD JOB! thanks!

PR #2006, opened 2021-05-27
fix: replace processEnv to env when no processEnv

zeniroy:
compilerOptions.processEnv replaced with env. but sometimes i want to use different env for compiling and running. 
so i changed compilerOptions.processEnv replaced with env when only processEnv not exist.


nchanged (approved):
that looks good and ready to go!

PR #2031, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/modules/buffer/index.js
+++ b/modules/buffer/index.js
@@ -765,7 +765,7 @@ function hexWrite(buf, string, offset, length) {
     length = strLen / 2;
   }
   for (var i = 0; i < length; ++i) {
-    var parsed = parseInt(string.substr(2 * i, 2), 16);
+    var parsed = parseInt(string.slice(2 * i, 2 * i + 2), 16);
     if (isNaN(parsed)) return i;
     buf[offset + i] = parsed;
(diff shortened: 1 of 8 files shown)


Havunen:
Why not just use String.prototype.substring ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

CommanderRoot:
substring() and slice() are very similar but [differ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) in some regards.
I personally prefer slice() over substring() as it's imo stricter. There is no "re-sorting" of arguments. It's also shorter and has an array counterpart which works the same way which makes it more intuative. slice() is also slightly faster but that probably doesn't matter that much.

